PR #21, opened 2023-08-28
STORY-905 Query Link edit component

goodguyry:
* Adds an editing interface for the Query Link block
  * Allows setting link text (default: "See All")
  * Allows overriding the link href (default: based on query context)
* Adds a Toolbar URL Selector component

https://alleyinteractive.atlassian.net/browse/STORY-905


renatonascalves (approved):
🌵

PR #22, opened 2023-08-28
STORY-906: query heading block

kaitlinbolling:
Adds a query heading block that inherits context from the parent query block and allows an override.

https://alleyinteractive.atlassian.net/browse/STORY-906


blocks/query-heading/edit.tsx
```diff
@@ -1,17 +1,119 @@
...
+import { addQueryArgs } from '@wordpress/url';
 
-import './index.scss';
+interface EditProps {
+  attributes: {
+    level?: number,
+    override?: string;
+  };
+  context?: {
+    curation?: any; // Shape TBD.
```
mogmarsh, line 23: did you want to flesh this out yet?

blocks/query-heading/render.php
```diff
@@ -2,36 +2,31 @@
...
-	if ( is_string( $wp_curate_heading ) && '' !== $wp_curate_heading ) :
-		?>
-			<h2 <?php echo wp_kses_data( get_block_wrapper_attributes( [ 'class' => 'wp-block-heading' ] ) ); ?>>
-				<?php echo esc_html( $wp_curate_heading ); ?>
-			</h2>
-		<?php
-	endif;
+if ( is_string( $wp_curate_heading ) && '' !== $wp_curate_heading ) :
+	?>
+		<<?php echo esc_attr( $wp_curate_heading_tag); ?> <?php echo wp_kses_data( get_block_wrapper_attributes( [ 'class' => 'wp-block-heading' ] ) ); ?>>
```
mogmarsh: does this class solve the concern of heading styles not getting loaded? I know you brought that up the other day...

src/class-wp-curate.php
```diff
@@ -12,4 +12,32 @@
  */
 class WP_Curate {
 
+    /**
```
mogmarsh: not in your PR, but we should change the comment on the Class.

src/class-wp-curate.php
```diff
@@ -12,4 +12,32 @@
...
+	 *
+	 * @param array<mixed> $curation Curation settings.
+	 * @return string Query heading.
+	 */
+	public static function get_query_heading( array $curation ): string {
+		$heading  = '';
+		$provider = $curation[ 'provider' ];
+
+		if ( taxonomy_exists( $provider ) && isset( $curation[ $provider ] ) && is_numeric( $curation[ $provider ] ) ) { // @phpstan-ignore-line
+			$term = get_term( $curation[ $provider ], $provider ); // @phpstan-ignore-line
```
mogmarsh, line 48: I was able to not ignore this line by running `intval()` on `$curation[ $provider ]`

mogmarsh, line 48:
```suggestion
			$term = get_term( intval( $curation[ $provider ] ), $provider );
```

mogmarsh (approved):
👍

src/class-wp-curate.php
kaitlinbolling: I'm still seeing an error: `Parameter #2 $taxonomy of function get_term expects string, mixed given.`. Should I not use `<mixed>` above?

blocks/query-heading/render.php
kaitlinbolling: No, I don't think it does. This was from The Wrap, but I actually don't think that class is needed.

blocks/query-heading/edit.tsx
kaitlinbolling: I'm not sure what the shape will be...I think I'm confused about how this context works.

PR #24, opened 2023-08-29
STORY-903: admin interface query

mogmarsh:
Sets up the admin interface for the Query block.


blocks/query/edit.tsx
```diff
@@ -1,17 +1,362 @@
...
+    postTypes?: string[];
+    posts?: any[];
+    terms?: {
+      [key: string]: any[];
+    };
+    searchTerm?: string;
+    query: {
+      [key: string]: string | number | number[] | string[];
+    }
+  };
```
kaitlinbolling, line 34: These attributes don't seem to match what's defined in block.json (min/max number of posts is missing, and block.json also has `pages`).

kaitlinbolling (approved):
👍 This all looks good to me! Just one question about attributes.

PR #28, opened 2023-08-31
STORY-911: Update to node 18

juliobranha:
https://alleyinteractive.atlassian.net/browse/STORY-911


kaitlinbolling (changes requested):
We should also update the node version in `package.json` [here](https://github.com/alleyinteractive/wp-curate/blob/develop/package.json#L7), and in the [.nvmrc](https://github.com/alleyinteractive/wp-curate/blob/develop/.nvmrc) file.

juliobranha:
> We should also update the node version in `package.json` [here](https://github.com/alleyinteractive/wp-curate/blob/develop/package.json#L7), and in the [.nvmrc](https://github.com/alleyinteractive/wp-curate/blob/develop/.nvmrc) file.

@kaitlinbolling Sorry!!! I did that but forgot to commit / push, fixed now

juliobranha:
> We should also update the node version in `package.json` [here](https://github.com/alleyinteractive/wp-curate/blob/develop/package.json#L7), and in the [.nvmrc](https://github.com/alleyinteractive/wp-curate/blob/develop/.nvmrc) file.

kaitlinbolling:
@juliobranha looking better! The build is failing though so there's a few more steps needed:

1. Some of the GitHub actions are defaulting to node 16. I'm not super familiar with GH actions, but we should be able to pass in node 18 via a `with` command. See https://github.com/alleyinteractive/.github#node. For example:
```
jobs:
  built-branch:
    uses: alleyinteractive/.github/.github/workflows/built-branch.yml@main
    with:
      node: 18
```
I think this is necessary for the built-branch, built-tag, and node-tests actions.

2. Node 18 is only compatible with NPM 9, so that will need to be updated in our package.json file [here](https://github.com/alleyinteractive/wp-curate/blob/develop/package.json#L8).
3. It looks like there might be some packages that require node 16 and/or npm 8, so we'll need to identify those, update them, and ensure there are no breaking changes.

kaitlinbolling (approved):
1️⃣ 8️⃣ 👍

PR #30, opened 2023-08-31
STORY-940: Condition block attributes interface should display depending on parent block

juliobranha:
https://alleyinteractive.atlassian.net/browse/STORY-940


blocks/condition/edit.tsx
```diff
@@ -91,41 +87,39 @@ export default function Edit({
...
-        { /* @ts-ignore */ }
-        <PanelBody
-          title={__('Index Condition', 'the-wrap')}
-        >
-          <p>{__('Checks the index of how many times the parent condition block has been rendered, ie "Equals to 0", "Greater than 5"', 'wp-curate')}</p>
+        { parentBlock === 'wp-curate/query' ? (
+          <PanelBody
+            title={__('Index Condition', 'wp-curate')}
+          >
+            <p>{__('Checks the index of how many times the parent condition block has been rendered, ie "Equals to 0", "Greater than 5"', 'wp-curate')}</p>
```
mogmarsh:
```suggestion
            <p>{__('Checks the index of how many times the parent condition block has been rendered, ie "Equal to 0", "Greater than 5"', 'wp-curate')}</p>
```

blocks/condition/edit.tsx
```diff
@@ -4,6 +4,8 @@ import {
   PanelBody, PanelRow, SelectControl, TextControl,
 } from '@wordpress/components';
 
+import useParentBlock from '@/hooks/use-parent-block';
```
mogmarsh:
```suggestion
import { useParentBlock } from '@alleyinteractive/block-editor-tools';
```

hooks/use-parent-block/index.tsx
mogmarsh, line 1: should be able to get rid of this file and use https://github.com/alleyinteractive/alley-scripts/tree/main/packages/block-editor-tools/src/hooks/use-parent-block
imported from https://www.npmjs.com/package/@alleyinteractive/block-editor-tools

mogmarsh (approved):
I recommend using `useParentBlock` from block-editor-tools, but otherwise 👍

PR #31, opened 2023-09-01
Support query deduplication

dlh01:


src/features/class-query-block-context.php
```diff
@@ -29,10 +53,7 @@ public function boot(): void {
 	 * @return array Updated context.
 	 */
 	public function filter_query_context( $context, $parsed_block ) {
-		$curated  = new Curated_Posts();
-		$registry = \WP_Block_Type_Registry::get_instance();
-
-		$block_type = $registry->get_registered( 'wp-curate/query' );
+		$block_type = $this->block_type_registry->get_registered( 'wp-curate/query' );
```
mogmarsh, line 56: should we support the core query too? 🤔

mogmarsh (approved):
👍

src/features/class-query-block-context.php
dlh01: It's possible. It might get wonky because the block attributes are different. But it might not. I'll noodle on it tonight.

PR #35, opened 2023-09-06
Feature/editor dedupe controls

mogmarsh:
brings over the block setting and post slotfill to enable/disable deduplication


blocks/query/block.json
```diff
@@ -20,50 +20,60 @@
...
-    "minNumberOfPosts": {
-      "type": "number",
-      "default": 1
+    "deduplication": {
+      "default": "inherit",
+      "enum": [
+        "inherit",
+        "never"
+      ],
+      "type": "string"
```
mogmarsh, line 29: this is the only new attribute. I just sorted the rest alphabetically.

dlh01 (approved):
🌳

PR #45, opened 2023-09-10
Refactor deduplication into smaller classes

dlh01:
Depends on https://github.com/alleyinteractive/wp-type-extensions/pull/11.


src/class-pinned-post-queries.php
```diff
@@ -0,0 +1,66 @@
...
+		$args['posts_per_page'] = $expected_per_page - \count( $pinned_post_ids );
+
+		$remaining       = new Exclude_Queries(
+			$this->pinned,
+			$args['posts_per_page'],
+			$this->origin,
+		);
+		$remaining_query = $remaining->query( $args );
+
+		return new Post_IDs_Query( array_merge( $pinned_post_ids, $remaining_query->post_ids() ) );
```
mogmarsh: I don't think we should fix this now, but I wanted to note that it is possible in the admin for users to pin posts in non-contiguous slots, for example:
`[ 2, , 6, , , 3 ]` or something like that. On the admin side, I'm looping through and inserting backfilled posts into the empty spots instead of doing an array merge, so we get `[ 2, 8, 6, 7, 5, 3 ]` instead of `[ 2, 6, 3, 8, 7, 5 ]` I'll create an issue for this to address later.

mogmarsh (approved):
looks good! 👍

src/class-pinned-post-queries.php
dlh01: Oooh, I hadn't noticed that. Thanks for mentioning it. That seems like a plain bug to me. I'll take a few minutes to try to address it before merging.

PR #53, opened 2023-09-13
fix/get query parent

mogmarsh:
reverting back to old code because we're not getting the direct parent, but the first ancestor that matches a block name (wp-curate/query)


kaitlinbolling (approved):
👍  I'm sorry I led you astray!

mogmarsh:
> 👍 I'm sorry I led you astray!

not your fault! it was a great idea and I didn't test thoroughly afterward.

PR #55, opened 2023-09-14
Use checkboxes for post types

### LINKED ISSUE #49: Change Post Type selector to checkboxes
### Description

In the Query block sidebar, there is a multiple select for the post types. Selecting or unselecting in a multiple select can be a hassle because you need to hold down the option key. We should switch this to use multiple checkboxes so the UI is intuitive and easier.

dlh01:
Fixes #49. 

Note that this component adds a decent amount of spacing between the label and the checkboxes. If that becomes an issue, it should be addressed in https://github.com/alleyinteractive/alley-scripts/, I think.


scottnelle (approved):
👻

Issue #56, opened 2023-09-14
`wp_curate_deduplication` meta is only registered for posts


mogmarsh:
### Description of the bug

The `wp_curate_deduplication` meta is only registered for posts, so it can't be enabled for any other post types (like pages).
https://github.com/alleyinteractive/wp-curate/blob/develop/config/post-meta.json#L3-L5

### Steps To Reproduce

1. Edit a page
2. Enable deduplication in the slot fill
3. Save
4. The toggle reverts back to unselected.

### Additional Information

We should either register it for all post types or move away from the json config file and register it manually using a filterable list of post types (which could be shared with https://github.com/alleyinteractive/wp-curate/issues/56)

renatonascalves:
@mogmarsh Is this issue still relevant?

The meta key is already registered by all post types, see here: https://github.com/alleyinteractive/wp-curate/blob/63587438c7bceba2feeab3bcd69abeb3f2296d9d/config/post-meta.json#L4

And I confirmed in a client project that it is saving properly in posts/pages...

mogmarsh:
@renatonascalves iI put in a quick fix to register it for all. https://github.com/alleyinteractive/wp-curate/pull/72 Will implement a better fix to register it only where necessary.

mogmarsh:
fixed in https://github.com/alleyinteractive/wp-curate/pull/72

Issue #54, opened 2023-09-14
Editor backfill query is hardcoded to posts


dlh01:
### Description of the bug

No matter the post types selected for the block, the `wp/v2/posts/` endpoint is used for searching, which returns only posts.

### Steps To Reproduce

1. Add `page` to the `wp_curate_duduplication_slotfill_post_types` filter.
2. Update the query block to draw backfill from only pages.
3. See posts display in the backfill instead of pages.

### Additional Information

See here: https://github.com/alleyinteractive/wp-curate/blob/b0d82733d8c0aebb698c16ff434c90ac742db71b/blocks/query/edit.tsx#L185.

### Acceptance Criteria

- Remove custom logic for posts endpoint https://github.com/alleyinteractive/wp-curate/blob/develop/src/features/class-rest-api.php
- Convert to using the search endpoint (if possible, even if the search endpoint needs to be augmented to include additional data for this type of query) or create a custom endpoint for WP Curate specifically.

mogmarsh:
we probably need to create a custom endpoint or use `/wp/v2/search`

mogmarsh:
ah yes, we had this fixed in the "first attempt": https://github.com/alleyinteractive/wp-curate/blob/first-attempt/src/class-wp-curate.php#L21-L35

mogmarsh:
update: this is fixed, but I don't know if it was fixed in the best way.

jacobschweitzer:
Working on re-factoring to use the `/search` endpoint.

mogmarsh:
@jacobschweitzer I don't think `/search` will work, because we're using the core Post Template block, which references a post type endpoint. See https://github.com/WordPress/gutenberg/blob/trunk/packages/block-library/src/post-template/edit.js#L210
I think we decided the best fix is to create a dummy custom post type called `wp-curate` so that `/wp/v2/wp-curate` works and have it return posts from all the selected post types, rather than modifying the behavior of `wp/v2/posts/

jacobschweitzer:
@mogmarsh I got a PR started though it needs a bit more work to get it across the finish line. I'm uncertain if this will be used in the project I'm on so I need to shift my focus back to other things for the time being.

kevinfodness:
@mogmarsh is this still relevant?

mogmarsh:
@kevinfodness yes and no... it works fine without it, but what we're doing feels a little "icky", so it's a nice to have, not a need to have at this point.
[...]

PR #59, opened 2023-09-15
add support for post_types to rest api

mogmarsh:
add support for post_types to rest api and allow `wp_curate_deduplication` on all post types.
handles https://github.com/alleyinteractive/wp-curate/issues/54 and sort of https://github.com/alleyinteractive/wp-curate/issues/56 (could be handled better)


src/features/class-rest-api.php
```diff
@@ -0,0 +1,43 @@
...
+	 * @param \WP_REST_Request                 $request The REST request.
+	 * @return array<array<int, string>|string>
+	 */
+	public function add_type_param( $query_args, $request ): array { // @phpstan-ignore-line
+		if ( ! empty( $request->get_param( 'type' ) ) && is_string( $request->get_param( 'type' ) ) ) {
+			$types                   = explode( ',', $request->get_param( 'type' ) );
+			$types                   = array_filter( $types, 'post_type_exists' );
+			$query_args['post_type'] = $types;
+		}
+		return $query_args;
```
dlh01, line 41: I get where this is coming from, but I'm concerned that it's not really the correct use of the REST API. It will cause the API to try to apply the schema for the `post` post type to posts from any post type, which might have unpredictable results. The `search` endpoint is more annoying to use but might be more correct for this case.

kaitlinbolling (approved):
👍 Looks like most of the changes in this PR were rolled back, but approving the small change to post-meta.json.

PR #60, opened 2023-09-15
Make sure context is provided when query block is an inner block

### LINKED ISSUE #58: Multiple Query blocks inside another block do not display correctly on the front end.
### Description of the bug

When there are multiple Query blocks inside a Column or Group block, the context on the front end seems to get confused or combined. All the Query blocks end up showing the same posts (same number of posts and same posts). Outside of the Group or Column block, they work as expected.

### Steps To Reproduce

1. Add a group block
2. Add multiple Query blocks. Set each one with a different number of posts, and maybe change the headline sizes so you can tell them apart.
3. Save and preview, you should see the same posts in both Query blocks and it will not match the number of posts selected for at least one of the blocks.

### Additional Information

_No response_

dlh01:
Fixes #58.

### DIFF
--- a/src/features/class-query-block-context.php
+++ b/src/features/class-query-block-context.php
@@ -18,6 +18,7 @@ use Alley\WP\Types\Post_Query;
 use Alley\WP\WP_Curate\Must_Include_Curated_Posts;
 use Alley\WP\WP_Curate\Plugin_Curated_Posts;
 use Alley\WP\WP_Curate\Recorded_Curated_Posts;
+use WP_Block;
 use WP_Block_Type;
 use WP_Block_Type_Registry;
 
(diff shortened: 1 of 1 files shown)


src/features/class-query-block-context.php
```diff
@@ -48,60 +49,81 @@ public function __construct(
...
-				qv: $this->stop_queries_var,
-				origin: new Plugin_Curated_Posts(
-					queries: new Variable_Post_Queries(
-						input: function () use ( $parsed_block ) {
-							$main_query = $this->main_query->query_object();
+		// Provide 'query' context to the plugin's query block.
+		if ( $plugin_block_type instanceof WP_Block_Type && $current_block->block_name() === $plugin_block_type->name ) {
+			$curated_posts = new Recorded_Curated_Posts(
+				// Deduplication is achieved by both recording to and excluding from the same place.
+				history: $this->history,
```
scottnelle, line 72: First time I've seen named arguments. I had to look at the docs to see what this was all about! TIL.

scottnelle (approved):
👻

src/features/class-query-block-context.php
dlh01: Yeah, I like them. With that and [constructor property promotion](https://github.com/alleyinteractive/wp-curate/blob/b0d82733d8c0aebb698c16ff434c90ac742db71b/src/class-recorded-curated-posts.php#L17-L26) (which I was skeptical of at first), you can make object initialization pretty clean and code-free.

PR #61, opened 2023-09-16
Rearrange query block sidebar panels

### LINKED ISSUE #50: Change order and label(s) of sidebar settings
### Description

Changing the pinned posts will probably happen more often than changing the backfill settings, so the pinned posts should be moved to the top. The section currently labeled "Block Settings" should be more descriptive, like "backfill settings".

dlh01:
Fixes #50, although I fully expect that the labels and positioning of these controls will continue to be refined.


kaitlinbolling (approved):
👍 Nice!

PR #84, opened 2023-10-24
Issue-73: In Query block add AND/OR option for multiple terms

### LINKED ISSUE #73: In Query block add AND/OR option for multiple terms
### Description

If multiple terms are selected, we should be able to select if the tax_query is AND or OR
This should be updated in the admin and the front end.

### Use Case

Users should have a drop down to choose AND or OR for the tax query behavior when multiple terms are entered.

mogmarsh:
## Description

This PR addresses issue #73: "In Query block add AND/OR option for multiple terms."

### Changes

- Allow users to select if the tax_query is AND or OR when multiple terms are selected
- Update the admin and front end to reflect the new functionality

### How to Test

1. Create a new Query block
2. Add multiple terms to the block
3. Use the new drop-down to choose between AND or OR for the tax query behavior

Fixes #73

### Summary by CodeRabbit

- New Feature: Enhanced the WP Curate plugin to version 1.2.0, introducing more control over post display with AND/OR operators in Query Parameters. This allows users to fine-tune their content curation based on specific taxonomies and term relations.
- Refactor: Simplified the codebase by introducing a new function `buildTermQueryArgs` that efficiently constructs term query arguments for the WP REST API.
- Bug Fix: Addressed an issue in version 1.1.0 that caused an error when the post type did not support meta. This ensures a smoother user experience across different post types.
- Documentation: Updated the CHANGELOG.md to reflect the new features and bug fixes in the recent versions.

### DIFF
--- a/blocks/query/edit.tsx
+++ b/blocks/query/edit.tsx
@@ -31,6 +31,8 @@ import {
   mainDedupe,
 } from '../../services/deduplicate';
 
+import buildTermQueryArgs from '../../services/buildTermQueryArgs';
+
 import './index.scss';
 
 interface Window {
@@ -72,6 +74,17 @@ export default function Edit({
     } = {},
   } = (window as any as Window);
 
+  const andOrOptions = [
+    {
+      label: __('AND', 'wp-curate'),
+      value: 'AND',
+    },
+    {
+      label: __('OR', 'wp-curate'),
+      value: 'OR',
+    },
+  ];
+
   // @ts-ignore
   const [isPostDeduplicating, postTypeObject] = useSelect(
     (select) => {
@@ -101,23 +114,13 @@ export default function Edit({
     return acc + hasTax;
   }, 0);
 
-  let termQueryArgs = '';
-  if (Object.keys(availableTaxonomies).length > 0) {
-    allowedTaxonomies.forEach((taxonomy) => {
-      if (terms[taxonomy]?.length > 0) {
-        const restBase = availableTaxonomies[taxonomy].rest_base;
-        if (restBase) {
-          termQueryArgs += `&${restBase}[terms]=${terms[taxonomy].map((term) => term.id).join(',')}`;
-          if (termRelations[taxonomy] !== '' && typeof termRelations[taxonomy] !== 'undefined') {
-            termQueryArgs += `&${restBase}[operator]=${termRelations[taxonomy]}`;
-          }
-        }
-      }
-    });
-    if (taxCount > 1) {
-      termQueryArgs += `&tax_relation=${taxRelation}`;
-    }
-  }
+  const termQueryArgs = buildTermQueryArgs(
+    allowedTaxonomies,
+    terms,
+    availableTaxonomies,
+    termRelations,
+    taxRelation,
+  );
 
   const manualPostIds = manualPosts.map((post) => (post ?? null)).join(',');
   const postTypeString = postTypes.join(',');
@@ -157,7 +160,7 @@ export default function Edit({
           per_page: 20,
         },
       );
-      path += termQueryArgs;
+      path += `&${termQueryArgs}`;
 
       apiFetch({
         path,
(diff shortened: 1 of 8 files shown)


blocks/query/edit.tsx
```diff
@@ -93,16 +96,27 @@ export default function Edit({
...
+          termQueryArgs += `&${restBase}[terms]=${terms[taxonomy].map((term) => term.id).join(',')}`;
+          if (termRelations[taxonomy] !== '' && typeof termRelations[taxonomy] !== 'undefined') {
+            termQueryArgs += `&${restBase}[operator]=${termRelations[taxonomy]}`;
+          }
         }
       }
     });
+    if (taxCount > 1) {
+      termQueryArgs += `&tax_relation=${taxRelation}`;
+    }
```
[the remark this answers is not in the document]
renatonascalves: I sort of agree with the premise of this suggestion. Meaning that there is quite a bit of logic into this file and that moving into a better place, with a better abstraction, will make future changes more manageable for future developers.

> Most modules have more users than developers, so it is better for the developers to suffer than the users.

> it is more important for a module to have a simple interface than a simple implementation.

blocks/query/edit.tsx
```diff
@@ -319,9 +341,49 @@ export default function Edit({
...
+                      {
+                        label: __('OR', 'wp-curate'),
+                        value: 'OR',
+                      },
+                    ]}
+                    onChange={(newValue) => setTermRelation(taxonomy, newValue)}
+                    value={termRelations[taxonomy]}
+                  />
+                ) : null}
+                <hr />
```
renatonascalves, line 356: Is this necessary if there is no terms? it was not available before.

blocks/query/edit.tsx
```diff
@@ -319,9 +341,49 @@
...
+                },
+                {
+                  label: __('OR', 'wp-curate'),
+                  value: 'OR',
+                },
+              ]}
+              onChange={(newValue) => setAttributes({ taxRelation: newValue })}
+              value={taxRelation}
+            />
+          ) : null }
```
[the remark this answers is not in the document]
renatonascalves: Agree! The same value is duplicated at the top.

blocks/query/edit.tsx
```diff
@@ -319,9 +341,49 @@ export default function Edit({
...
+                {terms[taxonomy]?.length > 1 ? (
+                  <SelectControl
+                    label={sprintf(
+                      __('%s Relation', 'wp-curate'),
+                      availableTaxonomies[taxonomy].name || taxonomy,
+                    )}
+                    help={__('AND: Posts must have all selected terms. OR: Posts may have one or more selected terms.', 'wp-curate')}
+                    options={[
+                      {
+                        label: __('AND', 'wp-curate'),
```
renatonascalves: Curious, if I needed to change this label: would that be possible?

wp-curate.php
```diff
@@ -3,7 +3,7 @@
  * Plugin Name: WP Curate
  * Plugin URI: https://github.com/alleyinteractive/wp-curate
  * Description: Plugin to curate homepages and other landing pages
- * Version: 1.1.0
+ * Version: 1.2.0
```
renatonascalves, line 6: https://github.com/alleyinteractive/wp-curate/blob/develop/CHANGELOG.md could use some update

renatonascalves (approved):
🌵

blocks/query/edit.tsx
mogmarsh: @renatonascalves I can probably move it to PHP and make it filterable, then expose it to JS with `wp_localize_script()`.
Do you think you will need to?

blocks/query/edit.tsx
mogmarsh: it should be showing if there is more than one term selected. It was using the default behavior before, this gives you more control. The default behavior is `OR` so I have updated that to be the default.

PR #90, opened 2023-10-30
Loading enhancement for blocks and slotfills

renatonascalves:
## Summary

- Enhance the loading logic to account for headless sites, where the post type is not available;
- Use `declare(strict_types=1);`;
- Improves phpDoc;
- `wp_curate_duduplication_slotfill_post_types` fallbacks to the default supported post types;

The previous approach didn't work very well with headless sites, where we need to access the context of the block. Mainly because `get_current_post_type` returns empty in those instances and there was no way of hooking into that.

This approach keeps the same logic but allows devs to load the block in headless context, if necessary, by using `wp_curate_load`.

## Summary by CodeRabbit

Refactor:
- Enhanced code quality by introducing strict types in various PHP files.
- Improved the logic for loading blocks and slotfills in the `wp-curate` package, now using the `load()` method of the `Supported_Post_Types` class.

New Features:
- Added two new filter hooks, `wp_curate_allowed_post_types` and `wp_curate_allowed_taxonomies`, providing more customization options.

Tests:
- Introduced a new test class `Supported_Post_Types_Test` to ensure the functionality of the `Supported_Post_Types` class.

Chores:
- Excluded the "tests/" directory from code analysis to focus on production code.


entries/slotfills/index.php
```diff
@@ -7,22 +7,25 @@
...
+
 namespace Alley\WP\WP_Curate;
 
 /**
  * Registers all slotfill assets so that they can be enqueued through Gutenberg in
  * the corresponding context.
  */
 function register_slotfills_scripts(): void {
+	$supported_post_types = new Supported_Post_Types();
+
```
attackant, line 20: Does it make sense to use a singleton for this?

src/class-supported-post-types.php
```diff
@@ -65,8 +97,11 @@ public function get_current_post_type(): string|false {
...
 				// phpcs:ignore WordPress.VIP.SuperGlobalInputUsage.AccessDetected, WordPress.Security.NonceVerification.Recommended
-				$post_id   = absint( $_GET['post'] );
-				$post_type = get_post_type( $post_id );
+				$post_type = get_post_type( absint( $_GET['post'] ) );
+
+				if ( ! $post_type ) {
+					$post_type = '';
+				}
 			// phpcs:ignore WordPress.VIP.SuperGlobalInputUsage.AccessDetected, WordPress.Security.NonceVerification.NoNonceVerification, WordPress.Security.NonceVerification.Recommended
 			} elseif ( 'post-new.php' === $pagenow ) {
```
attackant, line 106: Same question I asked on PR wp-curate/85 ... does it make sense/is it possible to use a nonce here instead of bypassing?

attackant (changes requested):
Asked a couple q's for review before explicit approval.

src/class-supported-post-types.php
```diff
@@ -50,12 +52,42 @@ public function get_supported_post_types(): array {
 		return apply_filters( 'wp_curate_supported_post_types', $this->supported_post_types );
 	}
 
+	/**
+	 * Load in the supported post types.
+	 *
+	 * Load a block or slotfill using WP Curate supported post types, or custon ones.
```
mogmarsh:
```suggestion
	 * Load a block or slotfill using WP Curate supported post types, or custom ones.
```

src/class-supported-post-types.php
mogmarsh: no, this is pulling from an admin url that does not have a nonce.

entries/slotfills/index.php
renatonascalves: It does!

PR #92, opened 2023-11-01
Feature/issue 73/add and or option multiple terms

mogmarsh:
Fixes issue in the PHP handling.

php tax_query wants `AND` or `IN`, but the REST API wants `AND` or `OR`

## Summary by CodeRabbit

Bug Fixes:
- Fixed a bug in the `WP Curate` plugin (version 1.4.2) that ensures the `operator` value for `tax_query` aligns with the requirements of both PHP and the REST API. This change improves the accuracy of post queries.
- Updated the `useSelect` hook in `blocks/query/edit.tsx` to return `isPostDeduplicating`, `postTypeObject`, and `postId`. This change enhances the post editing experience by excluding the current post from the query results.
- The `WP Curate` plugin version has been updated from 1.4.0 to 1.4.2, reflecting the above changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to `WP Curate` will be documented in this file.
 
+## 1.4.2 - 2023-11-01
+
+- Bug fix: PHP tax_query wants `AND` or `IN` for `operator`. REST API wants `AND` or `OR`.
+
 ## 1.4.0 - 2023-10-30
 
 - Bug fix: prevents error if `termRelations` attribute is not set.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to `WP Curate` will be documented in this file.
 ## 1.4.2 - 2023-11-01
 
 - Bug fix: PHP tax_query wants `AND` or `IN` for `operator`. REST API wants `AND` or `OR`.
+- Default operator should be `OR`/`IN`.
 
 ## 1.4.1 - 2023-11-01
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to `WP Curate` will be documented in this file.
 
+## 1.4.2 - 2023-11-01
+
+- Bug fix: PHP tax_query wants `AND` or `IN` for `operator`. REST API wants `AND` or `OR`.
```
renatonascalves, line 7: there is more here than just this change.

src/class-plugin-curated-posts.php
```diff
@@ -71,6 +72,7 @@ public function with_query_context( array $context, array $attributes, WP_Block_
 		if ( is_string( $search_term ) && strlen( $search_term ) > 0 ) {
 			$args['s'] = $search_term;
 		}
+		$args['post__not_in'] = [ get_the_id() ]; // phpcs:ignore WordPressVIPMinimum.Performance.WPQueryParams.PostNotIn_post__not_in
```
renatonascalves: What if I **want** to the current post to be available?

renatonascalves (approved):
🌵

src/class-plugin-curated-posts.php
mogmarsh: @renatonascalves I guess I could add a checkbox to allow it or something like that.

mogmarsh: maybe I should revert that part of the change and make a new issue to handle that. I really want to get this bug fix out.

renatonascalves: Got it! Feel free to move ahead. I was just picturing this would introduce changes that might affect other use cases.

PR #113, opened 2023-12-16
Set PHP and node versions on built release task

kevinfodness:
Fixes a bug where the built-release task fails because it's using the default version of node for that task, which is 18, when this project requires 20.

## Summary by CodeRabbit

- **Refactor**
  - Updated the build process to use the latest versions of Node.js and PHP for enhanced performance and compatibility.


mogmarsh (approved):
2️⃣ 0️⃣

vancoder (commented):
🌴

PR #123, opened 2024-01-25
Move PostPicker below InnerBlocks

### LINKED ISSUE #122: Post BlockControls Toolbar UI Overlays Post Selector "Pin" Button
### Description of the bug

When setting a grid layout in the Post Template, the Post BlockControl toolbar UI can potentially cover the "Pin a Post" button depending on the inner blocks selected. This can prevent a user from being able to access the "Pin a Post" button if there is only one inner block to select. The "Pin a Post" button should always be accessible and not obstructed by the Block Controls toolbar  to user regardless of inner blocks selected.

### Steps To Reproduce

1. Create a new Query block
2. Update the Post Template to use a grid layout of 2 or more columns
3. Select a post in the Query block to show the Post toolbar UI

### Additional Information

This issue is less of a problem when using the Outline view to select blocks. To a new user that might be more inclined to click directly on the blocks and not use Outline, it is more obvious and seems like a bug. 

2 column layout example with title inner block selected:
image

Suggestion, move the PostPicker below inner blocks and away from the BlockControls toolbar
image

image

ellm:
Avoid potential conflicts between BlockControl Toolbar and PostPicker button
See: https://github.com/alleyinteractive/wp-curate/issues/122

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to `WP Curate` will be documented in this file.
 
+## 1.5.1 - 2023-01-25
+
+- Bug Fix: Avoid BlockControl toolbar obstructing PostPicker button when Post inner blocks are selected.
+
 ## 1.5.0 - 2023-12-13
 
 - Enhancement: Bumps tested up to and requires WP to 6.4.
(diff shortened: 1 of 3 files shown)


mogmarsh (approved):
👍

PR #128, opened 2024-02-01
Rearrange Composer dependencies

dlh01:
- Require a specific version of Collision only during development, not when installing the plugin.
- Require a release version of Type Extensions instead of `main`.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,12 +24,12 @@
     "php": "^8.1",
     "alleyinteractive/composer-wordpress-autoloader": "^1.0",
     "alleyinteractive/traverse-reshape": "^2.0",
-    "alleyinteractive/wp-type-extensions": "dev-main",
-    "nunomaduro/collision": "^6.0"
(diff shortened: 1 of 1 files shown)


attackant (approved):
🐙

composer.json
```diff
@@ -24,12 +24,12 @@
...
     "alleyinteractive/composer-wordpress-autoloader": "^1.0",
     "alleyinteractive/traverse-reshape": "^2.0",
-    "alleyinteractive/wp-type-extensions": "dev-main",
-    "nunomaduro/collision": "^6.0"
+    "alleyinteractive/wp-type-extensions": "^1.0"
   },
   "require-dev": {
     "alleyinteractive/alley-coding-standards": "^2.0",
     "mantle-framework/testkit": "^0.12",
+    "nunomaduro/collision": "^6.0",
```
renatonascalves, line 32: `mantle-framework/testkit` should be already adding this package. So it is not necessary.

PR #131, opened 2024-02-08
Fix "Pin a Post" button margin

ellm:
This is a follow-up to #123.

Previously, the PostPicker appeared above the the InnerBlocks and some added margin was added below the buttons. Now that the buttons appear below, that margin should be applied to the top of the buttons container.


vancoder (approved):
🌴

Issue #133, opened 2024-02-14
Posts pinned twice in same Query block cause empty slot at the end


mogmarsh:
### Description of the bug

If a post is pinned twice within a Query block, it only gets pulled into the first slot leaving a blank spot at the end of the Query block

### Steps To Reproduce

Add a Query block
Pin the same post twice
You should see an empty slot at the end in both the editor and on the front end.

mogmarsh:
Proposed fix:
If the post list is currently `[1, 2, 3, 4]` and they select 4 in the first spot we would immediately remove it from the 4th spot leaving `[4, 2, 3, _]`
This would be handled in the block Edit function.

PR #134, opened 2024-02-15
Issue-133: Posts pinned twice in same Query block cause empty slot at the end

mogmarsh:
### Fixes Issue

Posts pinned twice in same Query block cause empty slot at the end. ([#133](https://github.com/alleyinteractive/wp-curate/issues/133))

### Description of the Change

- Implement a check in the Query block's edit function to handle the scenario where the same post is pinned more than once. This ensures that when a post is selected for a particular slot, it is immediately removed from any other slot it was previously occupying, thus preventing an empty slot from appearing at the end of the Query block. Additional improvements include cleaner code and null checks to ensure robust handling of edge cases.

### Steps to Reproduce the Issue

1. Add a Query block
2. Pin the same post twice
3. Observe an empty slot at the end in both the editor and on the front end.

### Proposed Fix

As suggested by [@mogmarsh](https://github.com/mogmarsh), the fix involves updating the block edit function. If the post list is currently `[1, 2, 3, 4]` and post `4` is selected for the first slot, it would immediately remove it from the 4th spot, resulting in the arrangement `[4, 2, 3, _]`. This change addresses the issue by ensuring that a post can only be pinned in one slot at a time. Further code cleanup and null checks have been implemented for enhanced stability and readability.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to `WP Curate` will be documented in this file.
 
+## 1.6.3 - 2024-02-14
+
+- Bug Fix: Selecting a post more than once in a Query block causes empty slots at the end.
+
 ## 1.6.2 - 2024-02-08
 
 - Bug Fix: Add intentional spacing before PostPicker buttons.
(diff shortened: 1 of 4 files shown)


renatonascalves (approved):
🗡️

Issue #142, opened 2024-03-01
Add option to pull Trending posts


mogmarsh:
### Description

The backfill should allow pulling featured posts from Parse.ly and eventually lede-beacon.

### Use Case

A user can select trending posts for the backfill source.

PR #143, opened 2024-03-01
Issue-142: Add option to pull Trending posts

mogmarsh:
### Description

The backfill should allow pulling featured posts from Parse.ly and eventually lede-beacon. We have now added the functionality to also pull trending posts, enhancing the variety of content that can be used for backfill. Editor controls have been set up to facilitate the selection of these posts, and a new endpoint has been developed to handle these queries effectively.

### Use Case

A user can select trending posts for the backfill source, in addition to the previously available options. This expansion allows for a more dynamic content selection process.

### Issue

[Add option to pull Trending posts](https://github.com/alleyinteractive/wp-curate/issues/142) - #142

### DIFF
--- a/blocks/query/block.json
+++ b/blocks/query/block.json
@@ -94,6 +94,10 @@
         "OR"
       ],
       "type": "string"
+    },
+    "trending": {
+      "default": false,
+      "type": "boolean"
     }
   }
 }
--- a/blocks/query/block.json
+++ b/blocks/query/block.json
@@ -95,9 +95,13 @@
       ],
       "type": "string"
     },
-    "trending": {
-      "default": false,
-      "type": "boolean"
+    "orderby": {
+      "default": "date",
+      "enum": [
+        "date",
+        "trending"
+      ],
+      "type": "string"
     }
   }
 }
--- a/blocks/query/edit.tsx
+++ b/blocks/query/edit.tsx
@@ -23,7 +23,6 @@ import {
 import { __, sprintf } from '@wordpress/i18n';
 import { addQueryArgs } from '@wordpress/url';
 
-import type { WP_REST_API_Post, WP_REST_API_Posts } from 'wp-types';
 import { Template } from '@wordpress/blocks';
 import type {
   EditProps,
--- a/blocks/query/edit.tsx
+++ b/blocks/query/edit.tsx
@@ -44,7 +44,7 @@ interface Window {
   wpCurateQueryBlock: {
     allowedPostTypes: Array<string>;
     allowedTaxonomies: Array<string>;
-    useParsely: string,
+    parselyAvailable: string,
   };
 }
 
@@ -78,7 +78,7 @@ export default function Edit({
     wpCurateQueryBlock: {
       allowedPostTypes = [],
       allowedTaxonomies = [],
-      useParsely = 'false',
+      parselyAvailable = 'false',
     } = {},
   } = (window as any as Window);
 
@@ -386,7 +386,7 @@ export default function Edit({
             onChange={(next) => setAttributes({ searchTerm: next })}
             value={searchTerm}
           />
-          { useParsely === 'true' ? (
+          { parselyAvailable === 'true' ? (
             <ToggleControl
               label={__('Show Trending Content from Parsely', 'wp-curate')}
               help={__('If enabled, the block will show trending content from Parsely.', 'wp-curate')}
--- a/blocks/query/edit.tsx
+++ b/blocks/query/edit.tsx
@@ -158,7 +158,7 @@ export default function Edit({
     }
     const fetchPosts = async () => {
       let path = addQueryArgs(
-        '/wp/v2/posts',
+        '/wp-curate/v1/posts',
         {
           search: debouncedSearchTerm,
           offset,
--- a/blocks/query/edit.tsx
+++ b/blocks/query/edit.tsx
@@ -11,6 +11,7 @@ import {
   RangeControl,
   SelectControl,
   TextControl,
+  ToggleControl,
 } from '@wordpress/components';
 import { useSelect } from '@wordpress/data';
 import {
@@ -44,6 +45,7 @@ interface Window {
   wpCurateQueryBlock: {
     allowedPostTypes: Array<string>;
     allowedTaxonomies: Array<string>;
+    useParsely: string,
   };
 }
 
@@ -69,6 +71,7 @@ export default function Edit({
     terms = {},
     termRelations = {},
     taxRelation = 'AND',
+    trending = false,
   },
   setAttributes,
 }: EditProps) {
@@ -76,6 +79,7 @@ export default function Edit({
     wpCurateQueryBlock: {
       allowedPostTypes = [],
       allowedTaxonomies = [],
+      useParsely = 'false',
     } = {},
   } = (window as any as Window);
 
@@ -161,6 +165,7 @@ export default function Edit({
           type: postTypeString,
           status: 'publish',
           per_page: 20,
+          trending,
         },
       );
       path += `&${termQueryArgs}`;
@@ -181,6 +186,7 @@ export default function Edit({
     offset,
     postTypeString,
     availableTaxonomies,
+    trending,
     setAttributes,
   ]);
 
@@ -384,6 +390,14 @@ export default function Edit({
             onChange={(next) => setAttributes({ searchTerm: next })}
             value={searchTerm}
           />
+          { useParsely === 'true' ? (
+            <ToggleControl
+              label={__('Show Trending Content from Parsely', 'wp-curate')}
+              help={__('If enabled, the block will show trending content from Parsely.', 'wp-curate')}
+              checked={trending}
+              onChange={(next) => setAttributes({ trending: next })}
+            />
+          ) : null }
         </PanelBody>
       </InspectorControls>
 
--- a/blocks/query/edit.tsx
+++ b/blocks/query/edit.tsx
@@ -173,10 +173,7 @@ export default function Edit({
       apiFetch({
         path,
       }).then((response) => {
-        const postIds: number[] = (response as WP_REST_API_Posts).map(
-          (post: WP_REST_API_Post) => post.id,
-        );
-        setAttributes({ backfillPosts: postIds });
+        setAttributes({ backfillPosts: response as Array<number> });
       });
     };
     fetchPosts();
--- a/blocks/query/edit.tsx
+++ b/blocks/query/edit.tsx
@@ -70,7 +70,7 @@ export default function Edit({
     terms = {},
     termRelations = {},
     taxRelation = 'AND',
-    trending = false,
+    orderby = 'date',
   },
   setAttributes,
 }: EditProps) {
@@ -164,7 +164,7 @@ export default function Edit({
           type: postTypeString,
           status: 'publish',
           per_page: 20,
-          trending,
+          orderby,
         },
       );
       path += `&${termQueryArgs}`;
@@ -182,7 +182,7 @@ export default function Edit({
     offset,
     postTypeString,
     availableTaxonomies,
-    trending,
+    orderby,
     setAttributes,
   ]);
 
@@ -390,8 +390,8 @@ export default function Edit({
             <ToggleControl
               label={__('Show Trending Content from Parsely', 'wp-curate')}
               help={__('If enabled, the block will show trending content from Parsely.', 'wp-curate')}
-              checked={trending}
-              onChange={(next) => setAttributes({ trending: next })}
+              checked={orderby === 'trending'}
+              onChange={(next) => setAttributes({ orderby: next ? 'trending' : 'date' })}
             />
           ) : null }
         </PanelBody>
--- a/blocks/query/index.php
+++ b/blocks/query/index.php
@@ -45,6 +45,7 @@ function wp_curate_query_block_init(): void {
 		[
 			'allowedPostTypes'  => $allowed_post_types,
 			'allowedTaxonomies' => $allowed_taxonomies,
+			'useParsely'        => apply_filters( 'wp_curate_use_parsely', false ) ? 'true' : 'false',
 		]
 	);
 }
--- a/blocks/query/index.php
+++ b/blocks/query/index.php
@@ -48,14 +48,14 @@ function wp_curate_query_block_init(): void {
 	 *
 	 * @param bool $use_parsely Whether to use Parsely.
 	 */
-	$use_parsley = apply_filters( 'wp_curate_use_parsely', false );
+	$parsely_available = apply_filters( 'wp_curate_use_parsely', false );
 	wp_localize_script(
 		'wp-curate-query-editor-script',
 		'wpCurateQueryBlock',
 		[
 			'allowedPostTypes'  => $allowed_post_types,
 			'allowedTaxonomies' => $allowed_taxonomies,
-			'useParsely'        => $use_parsley ? 'true' : 'false',
+			'parselyAvailable'  => $parsely_available ? 'true' : 'false',
 		]
 	);
 }
--- a/blocks/query/index.php
+++ b/blocks/query/index.php
@@ -31,21 +31,31 @@ function wp_curate_query_block_init(): void {
 
 	/**
 	 * Filter the post types that can be used in the Query block.
+	 *
+	 * @param array<string> $allowed_post_types The allowed post types.
 	 */
 	$allowed_post_types = apply_filters( 'wp_curate_allowed_post_types', [ 'post' ] );
 
 	/**
 	 * Filter the taxonomies that can be used in the Query block.
+	 *
+	 * @param array<string> $allowed_taxonomies The allowed taxonomies.
 	 */
 	$allowed_taxonomies = apply_filters( 'wp_curate_allowed_taxonomies', [ 'category', 'post_tag' ] );
 
+	/**
+	 * Filter whether to use Parsely.
+	 *
+	 * @param bool $use_parsely Whether to use Parsely.
+	 */
+	$use_parsley = apply_filters( 'wp_curate_use_parsely', false );
 	wp_localize_script(
 		'wp-curate-query-editor-script',
 		'wpCurateQueryBlock',
 		[
 			'allowedPostTypes'  => $allowed_post_types,
 			'allowedTaxonomies' => $allowed_taxonomies,
-			'useParsely'        => apply_filters( 'wp_curate_use_parsely', false ) ? 'true' : 'false',
+			'useParsely'        => $use_parsley ? 'true' : 'false',
 		]
 	);
 }
--- a/src/class-plugin-curated-posts.php
+++ b/src/class-plugin-curated-posts.php
@@ -43,7 +43,7 @@ final class Plugin_Curated_Posts implements Curated_Posts {
 			'no_found_rows'  => true,
 			'offset'         => $attributes['offset'] ?? $block_type->attributes['offset']['default'],
 			'order'          => 'DESC',
-			'orderby'        => $attributes['trending'] ? 'trending' : 'date',
+			'orderby'        => isset( $attributes['trending'] ) ? 'trending' : 'date',
 			'posts_per_page' => $attributes['numberOfPosts'] ?? $block_type->attributes['numberOfPosts']['default'],
 			'post_status'    => 'publish',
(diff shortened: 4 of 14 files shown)


blocks/query/block.json
```diff
@@ -94,6 +94,10 @@
         "OR"
       ],
       "type": "string"
+    },
+    "trending": {
+      "default": false,
+      "type": "boolean"
```
dlh01: Would it make sense to make `orderby` the attribute instead, with an enum of `"date", "trending"`? That could be forwarded to the query args and not need to be translated from `trending` to `orderby`.

blocks/query/index.php
```diff
@@ -45,6 +45,7 @@ function wp_curate_query_block_init(): void {
 		[
 			'allowedPostTypes'  => $allowed_post_types,
 			'allowedTaxonomies' => $allowed_taxonomies,
+			'useParsely'        => apply_filters( 'wp_curate_use_parsely', false ) ? 'true' : 'false',
```
dlh01: Filter needs docs.

src/class-plugin-curated-posts.php
```diff
@@ -43,7 +43,7 @@ public function with_query_context( array $context, array $attributes, WP_Block_
 			'no_found_rows'  => true,
 			'offset'         => $attributes['offset'] ?? $block_type->attributes['offset']['default'],
 			'order'          => 'DESC',
-			'orderby'        => 'date',
+			'orderby'        => isset( $attributes['trending'] ) ? 'trending' : 'date',
```
dlh01: What if `trending` is set but `false`?

src/class-trending-post-queries.php
```diff
@@ -0,0 +1,54 @@
...
+
+	/**
+	 * Query for posts using literal arguments.
+	 *
+	 * @param array<string, mixed> $args The arguments to be used in the query.
+	 * @return Post_Query
+	 */
+	public function query( array $args ): Post_Query {
+		if ( 'trending' === $args['orderby'] ) {
+			$parsely  = new Parsely_Support();
```
dlh01: I'd consider making this object a dependency in the constructor, unless there's a reason why the same instance can't be reused.

src/class-trending-post-queries.php
```diff
@@ -0,0 +1,54 @@
...
+	) {}
+
+	/**
+	 * Query for posts using literal arguments.
+	 *
+	 * @param array<string, mixed> $args The arguments to be used in the query.
+	 * @return Post_Query
+	 */
+	public function query( array $args ): Post_Query {
+		if ( 'trending' === $args['orderby'] ) {
```
dlh01:
```suggestion
		if ( isset( $args['orderby'] ) && 'trending' === $args['orderby'] ) {
```

src/class-trending-post-queries.php
```diff
@@ -0,0 +1,54 @@
...
+				return new WP_Query_Envelope(
+					new \WP_Query(
+						[
+							'post__in'            => $trending,
+							'post_type'           => 'any',
+							'ignore_sticky_posts' => true,
+							'orderby'             => 'post__in',
+						]
+					)
+				);
```
dlh01:
```suggestion
				return new Post_IDs_Query( $trending );
```

You'd need to have a `use` statement for that class, too.

src/features/class-parsely-support.php
```diff
@@ -0,0 +1,141 @@
...
+	 *
+	 * @param array<number>        $posts The posts, which should be an empty array.
+	 * @param array<string, mixed> $args The WP_Query args.
+	 * @return array<number> Array of post IDs.
+	 */
+	public function add_parsely_trending_posts_query( array $posts, array $args ): array {
+		$parsely = $GLOBALS['parsely'];
+		if ( ! $parsely->api_secret_is_set() ) {
+			return $posts;
+		}
```
dlh01, line 51: If this is going to be checked again here (which seems sensible), why worry about also checking it in `boot()`?

src/features/class-parsely-support.php
```diff
@@ -0,0 +1,141 @@
...
+	 * @param array<string, mixed> $args The WP_Query args.
+	 * @return array<number> An array of post IDs.
+	 */
+	public function get_trending_posts( array $args ): array {
+		// TODO: Add failover if we're not on production.
+		/**
+		 * Filter the period start for the Parsely API.
+		 *
+		 * @param string $period_start The period start.
+		 * @return string The period start.
```
dlh01: Filters don't get a `@return` in the docs, per https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#4-hooks-actions-and-filters.

src/features/class-parsely-support.php
```diff
@@ -0,0 +1,141 @@
...
+		 */
+		$period_start = apply_filters( 'wp_curate_parsely_period_start', '1d' );
+		$parsely_args = [
+			'limit'        => $args['posts_per_page'],
```
dlh01: What if this isn't included in `$args`?

src/features/class-parsely-support.php
```diff
@@ -0,0 +1,141 @@
...
+			'sort'         => 'views',
+			'period_start' => $period_start,
+			'period_end'   => 'now',
+		];
+		if ( isset( $args['tax_query'] ) && is_array( $args['tax_query'] ) ) {
+			foreach ( $args['tax_query'] as $tax_query ) {
+				if ( isset( $tax_query['taxonomy'] ) && 'category' === $tax_query['taxonomy'] ) {
+					$parsely_args['section'] = implode( ', ', $this->get_slugs_from_term_ids( $tax_query['terms'], $tax_query['taxonomy'] ) );
```
dlh01, line 94: Is it hardcoded in the Parsely plugin that `section` corresponds to `category`? Could it correspond to other taxonomies? 

Also, does this implementation assume that the official Parsely plugin is being used and not a custom integration? That seems fair to assume, but it's worth documenting.

src/features/class-parsely-support.php
```diff
@@ -0,0 +1,141 @@
...
+	}
+
+	/**
+	 * Gets the trending posts from Parsely.
+	 *
+	 * @param array<string, mixed> $args The WP_Query args.
+	 * @return array<number> An array of post IDs.
+	 */
+	public function get_trending_posts( array $args ): array {
+		// TODO: Add failover if we're not on production.
```
dlh01: What's the plan for this TODO?

src/features/class-parsely-support.php
```diff
@@ -0,0 +1,141 @@
...
+		$ids       = wp_cache_get( $cache_key );
+		if ( false === $ids ) {
+			$api   = new Analytics_Posts_API( $GLOBALS['parsely'] ); // @phpstan-ignore-line
+			$posts = $api->get_posts_analytics( $parsely_args ); // @phpstan-ignore-line
+			$ids   = array_map(
+				function ( $post ) {
+					// Check if the metadata contains post_id, if not, use the URL to get the post ID.
+					$metadata = json_decode( $post['metadata'] ?? '', true );
+					if ( is_array( $metadata ) && ! empty( $metadata ) && isset( $metadata['post_id'] ) ) {
+						$post_id = intval( $metadata['post_id'] );
```
dlh01:
```suggestion
						$post_id = (int) $metadata['post_id'];
```

src/features/class-parsely-support.php
```diff
@@ -0,0 +1,141 @@
...
+					} elseif ( function_exists( 'wpcom_vip_url_to_postid' ) ) {
+							$post_id = wpcom_vip_url_to_postid( $post['url'] );
```
dlh01:
```suggestion
						$post_id = wpcom_vip_url_to_postid( $post['url'] );
```

src/features/class-parsely-support.php
```diff
@@ -0,0 +1,141 @@
...
+					} else {
+						$post_id = url_to_postid( $post['url'] ); // phpcs:ignore WordPressVIPMinimum.Functions.RestrictedFunctions.url_to_postid_url_to_postid
+					}
+					/**
+					 * Filters the post ID derived from Parsely post object.
+					 *
+					 * @param int $post_id The post ID.
+					 * @param array $post The Parsely post object.
+					 * @return int The post ID.
```
dlh01:
```suggestion
```

src/features/class-parsely-support.php
dlh01: Similar thought about the `post_id` key here with respect to the official Parsely plugin.

src/features/class-parsely-support.php
```diff
@@ -0,0 +1,141 @@
...
+					 */
+					return apply_filters( 'wp_curate_parsely_post_to_post_id', $post_id, $post );
+				},
+				$posts
+			);
+			wp_cache_set( $cache_key, $ids, '', 10 * MINUTE_IN_SECONDS );
+		}
+		$ids = array_map( 'intval', $ids ); // @phpstan-ignore-line
```
dlh01: What's this `ignore` for?

src/features/class-parsely-support.php
```diff
@@ -0,0 +1,141 @@
...
+
+		return( $ids );
```
dlh01:
```suggestion
		return $ids;
```

src/features/class-rest-api.php
```diff
@@ -23,26 +23,127 @@ public function __construct() {}
...
+	 *
+	 * @param WP_REST_Request $request The request object.
+	 * @return array<int> The post IDs.
+	 */
+	public function get_posts( WP_REST_Request $request ): array { // phpcs:ignore Squiz.Functions.MultiLineFunctionDeclaration.ContentAfterBrace @phpstan-ignore-line
+		$search_term      = $request->get_param( 'search' ) ?? '';
+		$offset           = $request->get_param( 'offset' ) ?? 0;
+		$post_type_string = $request->get_param( 'post_type' ) ?? 'post';
+		$per_page         = $request->get_param( 'per_page' ) ?? 20;
+		$trending         = 'true' === $request->get_param( 'trending' ) ? true : false;
```
dlh01:
```suggestion
		$trending         = 'true' === $request->get_param( 'trending' );
```

src/features/class-rest-api.php
```diff
@@ -23,26 +23,127 @@ public function __construct() {}
...
+
+		if ( ! is_string( $post_type_string ) ) {
+			$post_type_string = 'post';
+		}
+
+		/**
+		 * Filters the allowed taxonomies.
+		 *
+		 * @param array<string> $allowed_taxonomies The allowed taxonomies.
+		 * @return array<string> The allowed taxonomies.
```
dlh01:
```suggestion
```

This could alternatively use the core standard for duplicate hook docs: https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#4-hooks-actions-and-filters

src/features/class-rest-api.php
```diff
@@ -23,26 +23,127 @@ public function __construct() {}
...
+		 */
+		$allowed_taxonomies = apply_filters( 'wp_curate_allowed_taxonomies', [ 'category', 'post_tag' ] );
+		$taxonomies         = array_map( 'get_taxonomy', $allowed_taxonomies );
+		$taxonomies         = array_filter( $taxonomies, 'is_object' );
```
dlh01, line 71: Could flip the order here to `array_filter` by `taxonomy_exists()` and then map to `get_taxonomy()`.

src/features/class-rest-api.php
```diff
@@ -23,26 +23,127 @@ public function __construct() {}
...
+		}
+
+		$post_types = explode( ',', $post_type_string );
+		/**
+		 * Filters the allowed post types.
+		 *
+		 * @param array<string> $allowed_post_types The allowed post types.
+		 * @return array<string> The allowed post types.
+		 */
+		$allowed_post_types = apply_filters( 'wp_curate_allowed_post_types', [ 'post' ] );
```
dlh01, line 107: As above with the hook docs.

src/features/class-rest-api.php
```diff
@@ -23,26 +23,127 @@ public function __construct() {}
...
+			 * Filters the trending posts query.
+			 *
+			 * @param array<string, mixed> $posts The posts.
+			 * @param array<string, mixed> $args The arguments.
+			 */
+			$posts = apply_filters( 'wp_curate_trending_posts_query', [], $args );
+		}
+		if ( empty( $posts ) ) {
+			$query = new \WP_Query( $args );
+			$posts = $query->posts;
```
dlh01, line 141: Am I following that the "correct" results of this endpoint would be the same results that are going to appear on the frontend for the block? Would we want to use the same `Post_Queries` instance that are used when filtering the block output here instead of a plain `WP_Query`?

dlh01 (commented):
High-level it's looking good. I've got some questions on the details.

blocks/query/block.json
mogmarsh: yes, that was my first instinct but wasn't sure about adding things to query args. I'll change this.

src/features/class-parsely-support.php
mogmarsh: @dlh01 I want to conditionally run 
```
add_filter( 'wp_curate_use_parsely', '__return_true' );
```
only if the secret is set, which I'm doing in `boot()`

src/features/class-parsely-support.php
mogmarsh: phpstan doesn't think `'intval'` is a callable

src/features/class-rest-api.php
mogmarsh: @dlh01 the "correct" results will not have any pinned posts or deduplication. That all happens client side in the editor, so it can be just a plain `WP_Query` or the `Post_Queries` equivalent of a plain `WP_Query`. We do want to support the trending override, of course. So we could do just `Trending_Post_Queries` and `Variable_Post_Queries`.

src/features/class-parsely-support.php
mogmarsh: I have added some logic to handle Parsely options better.

we check for the `Parsely\Parsely` class, so it won't work without the plugin installed. I'll need to write some docs on this feature at some point.

src/features/class-rest-api.php
mogmarsh: had to flip them back to this to make phpstan happy.

blocks/query/edit.tsx
```diff
@@ -44,6 +44,7 @@ interface Window {
   wpCurateQueryBlock: {
     allowedPostTypes: Array<string>;
     allowedTaxonomies: Array<string>;
+    useParsely: string,
```
kevinfodness: Is this really to determine whether Parse.ly support is _available_, rather than whether the query should use Parse.ly to pull its data (vs. a term-based query)? If so, I'd recommend naming this key something like `parselyAvailable`

src/class-trending-post-queries.php
```diff
@@ -0,0 +1,47 @@
...
+use Alley\WP\Types\Post_Queries;
+use Alley\WP\Types\Post_Query;
+use Alley\WP\Post_Query\WP_Query_Envelope;
+use Alley\WP\WP_Curate\Features\Parsely_Support;
+use Alley\WP\Post_Query\Post_IDs_Query;
+
+/**
+ * Pull trending posts from Parsely.
+ */
+final class Trending_Post_Queries implements Post_Queries {
```
kevinfodness, line 19: If this is specific to Parse.ly, then it should be named as such. If it's intended to be a generic trending posts class, then it should be structured so that it supports multiple providers, of which Parse.ly will be the first (and we could add other providers like GA4 later).

src/features/class-parsely-support.php
```diff
@@ -0,0 +1,152 @@
...
+	 * @return array<int> An array of post IDs.
+	 */
+	public function get_trending_posts( array $args ): array {
+		$parsely_options = $GLOBALS['parsely']->get_options();
+		/**
+		 * Filter the period start for the Parsely API.
+		 *
+		 * @param string $period_start The period start.
+		 */
+		$period_start = apply_filters( 'wp_curate_parsely_period_start', '1d' );
```
kevinfodness: Could pass `$args` here too, so implementers could make different decisions about the period based on the arguments passed to the function.

src/features/class-parsely-support.php
```diff
@@ -0,0 +1,152 @@
...
+		$parsely_args = [
+			'limit'        => $args['posts_per_page'] ?? 20,
```
kevinfodness: Could fall back to global posts per page setting here instead of hardcoding 20.

src/features/class-parsely-support.php
```diff
@@ -0,0 +1,152 @@
...
+			'sort'         => 'views',
+			'period_start' => $period_start,
+			'period_end'   => 'now',
```
kevinfodness: Might as well make this value filterable too.

src/features/class-parsely-support.php
```diff
@@ -0,0 +1,152 @@
...
+				$tags                = explode( ', ', $parsely_args['tag'] ?? '' );
+				$sections            = explode( ', ', $parsely_args['section'] ?? '' );
+				$parsely_args['tag'] = implode( ', ', array_merge( $tags, $sections ) );
+			}
+		}
+		$cache_key = 'parsely_trending_posts_' . md5( wp_json_encode( $parsely_args ) ); // @phpstan-ignore-line - wp_Json_encode not likely to return false.
+		$ids       = wp_cache_get( $cache_key );
+		if ( false === $ids || ! is_array( $ids ) ) {
+			$api   = new Analytics_Posts_API( $GLOBALS['parsely'] ); // @phpstan-ignore-line
+			$posts = $api->get_posts_analytics( $parsely_args ); // @phpstan-ignore-line
```
kevinfodness: Probably PHPStan is complaining here that these functions might fail - imo we should not suppress PHPStan here, and should instead bail out if there is a failure.

src/features/class-parsely-support.php
```diff
@@ -0,0 +1,152 @@
...
+			$ids   = array_map(
+				function ( $post ) {
+					// Check if the metadata contains post_id, if not, use the URL to get the post ID.
+					$metadata = json_decode( $post['metadata'] ?? '', true );
+					if ( is_array( $metadata ) && ! empty( $metadata ) && isset( $metadata['post_id'] ) ) {
```
kevinfodness: Empty check not needed here:
```suggestion
					if ( is_array( $metadata ) && isset( $metadata['post_id'] ) ) {
```

src/features/class-parsely-support.php
```diff
@@ -0,0 +1,152 @@
...
+					 * Filters the post ID derived from Parsely post object.
+					 *
+					 * @param int $post_id The post ID.
+					 * @param array $post The Parsely post object.
+					 */
+					return apply_filters( 'wp_curate_parsely_post_to_post_id', $post_id, $post );
+				},
+				$posts
+			);
+			wp_cache_set( $cache_key, $ids, '', 10 * MINUTE_IN_SECONDS );
```
kevinfodness: Cache lifetime could also be filterable.

src/features/class-rest-api.php
```diff
@@ -23,26 +23,125 @@ public function __construct() {}
...
-			$query_args['post_type'] = $types;
+	public function register_endpoints(): void {
+		register_rest_route(
+			'wp-curate/v1',
+			'/posts/',
+			[
+				'methods'             => 'GET',
+				'callback'            => [ $this, 'get_posts' ],
+				'permission_callback' => function () {
+					return current_user_can( 'edit_posts' );
```
kevinfodness: I wonder if this is too restrictive ... what if a user doesn't have the ability to edit the core posts post type but does have the ability to edit a custom post type? I don't think we're exposing any sensitive data here that wouldn't otherwise be available by the main posts endpoint, which is public, so maybe we make this a simple "is user logged in" check instead?

kevinfodness (dismissed):

blocks/query/edit.tsx
mogmarsh: yes, that is correct. good call.

src/class-trending-post-queries.php
mogmarsh: @kevinfodness supporting multiple providers was my plan... I kind of figured we would sort out how structure it when we added the second provider.

kevinfodness: Ah, we're using named arguments instead of positional ones, this works for me!

attackant (approved):
Looks good but feel free to keep getting others feedback on this if you need to 🐙

PR #149, opened 2024-03-13
Bug Fix: Query block not updating with custom post types

ellm:
- Fixes missing param value `Rest_Api::get_post` request
- Fixes lost support for custom post types when building query parameters in Query block

### DIFF
--- a/src/features/class-rest-api.php
+++ b/src/features/class-rest-api.php
@@ -24,6 +24,7 @@ final class Rest_Api implements Feature {
 	 */
 	public function boot(): void {
 		add_action( 'rest_api_init', [ $this, 'register_endpoints' ] );
+		add_filter( 'rest_post_query', [ $this, 'add_type_param' ], 10, 2 );
 	}
 
 	/**
@@ -141,5 +142,25 @@ final class Rest_Api implements Feature {
 			$posts = $query->posts;
 		}
 		return array_map( 'intval', $posts ); // @phpstan-ignore-line
+	}
+
+	/**
+	 * Add post_type to rest post query if the type param is set.
+	 *
+	 * @param array<array<int, string>|string> $query_args The existing query args.
+	 * @param WP_REST_Request                  $request The REST request.
+	 * @return array<array<int, string>|string>
+	 */
+	// @phpstan-ignore-next-line
+	public function add_type_param( $query_args, $request ): array { // phpcs:ignore Squiz.Commenting.FunctionComment.WrongStyle
+		$type = $request->get_param( 'type' );
+
+		if ( ! empty( $type ) && is_string( $type ) ) {
+			$types                   = explode( ',', $type );
+			$types                   = array_filter( $types, 'post_type_exists' );
+			$query_args['post_type'] = $types;
+		}
+
+		return $query_args;
 	}
 }
--- a/src/features/class-rest-api.php
+++ b/src/features/class-rest-api.php
@@ -147,12 +147,23 @@ final class Rest_Api implements Feature {
 	/**
 	 * Add post_type to rest post query if the type param is set.
 	 *
+	 * NOTE: This is a temporary solution that will be replaced by a more robust solution in the future.
+	 *
 	 * @param array<array<int, string>|string> $query_args The existing query args.
 	 * @param WP_REST_Request                  $request The REST request.
 	 * @return array<array<int, string>|string>
 	 */
 	// @phpstan-ignore-next-line
 	public function add_type_param( $query_args, $request ): array { // phpcs:ignore Squiz.Commenting.FunctionComment.WrongStyle
+		// Check if the user is logged in.
+		if ( ! \is_user_logged_in() ) {
+			return $query_args;
+		}
+		// Check the context.
+		if ( 'edit' !== $request['context'] ) {
+			return $query_args;
+		}
+
 		$type = $request->get_param( 'type' );
 
 		if ( ! empty( $type ) && is_string( $type ) ) {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to `WP Curate` will be documented in this file.
 
+## 1.7.1 - 2024-03-13
+
+- Bug Fix: Query block does not update with posts from custom post types when selected in Query Paramaters block settings.
+
 ## 1.7.0 - 2024-03-06
 
 - Enhancement: Integration with [Parse.ly plugin](https://wordpress.org/plugins/wp-parsely/) to support querying trending posts.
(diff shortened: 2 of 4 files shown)


src/features/class-rest-api.php
```diff
@@ -24,6 +24,7 @@ public function __construct() {}
 	 */
 	public function boot(): void {
 		add_action( 'rest_api_init', [ $this, 'register_endpoints' ] );
+		add_filter( 'rest_post_query', [ $this, 'add_type_param' ], 10, 2 );
```
mboynes, line 27: I'm a little iffy on this update.

First, to summarize my understanding of this: we're hooking into the filter `rest_{$post_type}_query`, and we're adding the ability for REST requests to the endpoints for the _post_ post type to allow those endpoints to return content from _any_ post type.

If my understanding is accurate...

1. Could this potentially leak private data? If I add `?type=revision`, am I able to load up revisions at the post endpoints? What about other private post types and those that have explicitly disabled REST support?
2. This is a significant change to core WP endpoints and I could imagine edge-case conflicts, especially with a generic argument like "type".
3. Philosophically, it feels _wrong_ to modify the endpoint for one resource to return another resource.

Perhaps this is something that warrants a custom endpoint?

mboynes (dismissed):
I blocked this given the potential security issue. This might warrant some discussion, feel free to bring it to `#opensource` to bounce ideas!

src/features/class-rest-api.php
ellm: Thanks @mboynes - these are all great points. I need to think on this some more but looping in @mogmarsh here as well. We were using this filter previously with the same callback function. It was removed as part of the last v1.7.0 release

mogmarsh: @mboynes we're using the core Post Template block which expects one Post Type to be set in the parent core Query Loop block, and pulls posts from that one post type https://github.com/WordPress/gutenberg/blob/trunk/packages/block-library/src/post-template/edit.js#L184-L190
We're allowing multiple post types, so we need someway to fetch posts of multiple types from one endpoint. This is only necessary in the editor, so maybe we could check if the user is logged in. 
A custom endpoint might work, but we would have to put it in the core namespace, so its url was `/wp-json/wp/v2/<something>` to be able to work with the existing core code.
I'm thinking the login check might be the way to go, but I'm open to suggestions and discussion.

mslinnea: I think a custom endpoint would be cleanest. The post types available should be restricted by the user's capabilities.  There are post types used for logs, feedback forms, etc, which may be restricted to some user roles.

dlh01: (3) is the overriding point, to me. Although I understand the tension between what the plugin is trying to do and the design of the REST API, mixing post types isn't the way the API is meant to be used.

Out of the client sites that have implemented this plugin so far, how many instances have there been of query blocks that pull in content from multiple post types? Is there a real-world need for this feature right now?

kevinfodness: Based on the link that @mogmarsh provided, core is using getEntityRecord to fetch results for the post template, which is why we went down this road in the first place. I'm not sure if it's possible to use the Post Template block and override where it gets its data from for backfill, but that's the next place we should investigate, imo. I'm in agreement with everything said here - if we can create a custom endpoint to serve this data, that's ideal. If it's truly impossible, and we have to use a hack of some sort, then we can add some security measures to prevent it from being abused (e.g., check to ensure the user is logged in, only allow public post types (which revisions aren't), check referer to ensure it's coming from the WP admin, etc).

kevinfodness: To your point @dlh01 I'm honestly not sure. That said, it's a killer feature for this plugin to have, because core forces you to pick a single post type. If, for example, there are multiple post types that all put content in the same category and you want to have a category list that shows all posts across multiple post types that are within that category, you can't do that with core's Query Loop block, but you should be able to do that with WP Curate. However, if we aren't aware of any sites that are currently using this feature, then there's not pressure to fix it quickly and we can find the right solution to this problem.

srtfisher: A custom endpoint seems like the way to go to echo all of you. Mixing different post types on a core post type controller seems wrong.

mogmarsh: @kevinfodness I just want to clarify that there may not be pressure to fix multiple post types in one Query block, but we do currently have a bug where Post is hard-coded, so other post types do not work in the editor right now. I just ran into this. So we **do** have pressure to fix that bug. Whether we fix it by hacking the core post endpoint with some security or we fix it by removing the multiple post type support until we can figure out a better fix doesn't really matter to me.

mboynes: Personally, I'd favor a custom endpoint in the core wp namespace over manipulating core endpoints, recognizing that that's not ideal _either_.

mogmarsh: I'll see if I can pass a post type of `wp-curate` and if it attempts to hit `/wp-json/wp/v2/wp-curate`. If so, that might be an option. It may be that the post type has to be registered, though.

mogmarsh: this does not work... it either validates the post type or uses the post type to look up the endpoint.

emilyatmobtown: FWIW if this comes down to a constraint in the Post Template block, that has been an obstacle in our use of WPC from the start. It's a great tool, but more than often than not has limited our application of WPC. Perhaps we should explore making our own version?

dlh01: Since we've established that this problem only affects the editor, and since it's at least uncommon for users to select multiple post types, can we work around the problem for now by having the editor show results from only a single post type, even if multiple are selected? Maybe it always shows `post` if that's selected, otherwise the first post type from the list. A warning could be displayed about this limitation at the same time.

>  it's a killer feature for this plugin to have, because core forces you to pick a single post type. If, for example, there are multiple post types that all put content in the same category and you want to have a category list that shows all posts across multiple post types that are within that category, you can't do that with core's Query Loop block, but you should be able to do that with WP Curate.

On the other hand, what if I pick multiple post types and a category, but not all of my selected post types support categories? As a user, I might expect that only those post types that support categories will be filtered by the category, but I think the actual behavior is that the post types without category support are excluded. Describing that cause-and-effect to the user is a UX challenge, at least.

Or, what if I pick zero post types? That's currently possible, and maybe it makes sense because someone might be trying to emphasize that the block is being used for only curated posts. But the actual behavior is to fall back to posts anyway.

I don't dispute at all that it's a killer feature, but I do think there's a fair amount to weigh against the impression that demand for the feature currently seems to be low.

emilyatmobtown: I can't speak to wider demand, but it's a high-priority blocker on one project that's already tightly tied to WPC.

mogmarsh: We discussed this today and agreed to this approach (with a few checks) as a short term fix, then we will implement a better fix soon.

mboynes (approved):

PR #152, opened 2024-03-14
Add support for GraphQL via WPGraphQL hooks

### LINKED ISSUE #150: Add support for GraphQL via WPGraphQL hooks
### Description

I would incorporate support for WPGraphQL using it's custom hooks so WP Curate can be used with GraphQL 

### Use Case

When a user installs WP Curate it should provide support for the WPGraphQL plugin with helpful additions for creating custom GraphQL queries

ellm:


src/features/class-graphql.php
```diff
@@ -0,0 +1,141 @@
...
+	private $allowed_post_types;
+
+	/**
+	 * Get the GraphQL Type by Post Type.
+	 *
+	 * @param  string $post_type_string Name of the post type. Ex: 'post'.
+	 *
+	 * @return string GraphQL Type name. Ex: 'Post'.
+	 */
+	private function get_graphql_type_by_post_type( $post_type_string ): string {
```
kevinfodness: Add type to function parameter also:
```suggestion
	private function get_graphql_type_by_post_type( string $post_type_string ): string {
```

src/features/class-graphql.php
```diff
@@ -0,0 +1,141 @@
...
+		$post_type_object = get_post_type_object( $post_type_string );
+
+		if ( empty( $post_type_object ) ) {
+			return '';
+		}
+
+		// Only return a GraphQL type for allowed post types in WP Curate.
+		if ( ! in_array( $post_type_string, $this->allowed_post_types, true ) ) {
```
kevinfodness, line 37: This check should come first, because it doesn't rely on fetching data from somewhere else (e.g., get_post_type_object).

src/features/class-graphql.php
```diff
@@ -0,0 +1,141 @@
...
+		}
+
+		return $interface_to_types;
+	}
+
+	/**
+	 * Set up.
+	 */
+	public function __construct() {
+		$this->allowed_post_types = apply_filters( 'wp_curate_allowed_post_types', [ 'post', 'opinion' ] );
```
kevinfodness: Remove `opinion` here, that's a custom post type that we can't expect to exist. The call to this filter should mirror the other ones in the plugin, which default to only `post`:
```suggestion
		$this->allowed_post_types = apply_filters( 'wp_curate_allowed_post_types', [ 'post' ] );
```

renatonascalves: The `opinion` post type looks client specific, no?

src/features/class-graphql.php
```diff
@@ -0,0 +1,141 @@
...
+	}
+
+	/**
+	 * Boot the feature.
+	 */
+	public function boot(): void {
+		// Assumes that WPGraphQL has been installed as a composer dependency of a parent project.
+		if ( class_exists( 'WPGraphQL' ) ) {
+			add_action( 'graphql_register_types', [ $this, 'graphql_register_types' ] );
+		}
```
kevinfodness: The way the hook system works means that we don't actually need to see if the WPGraphQL class exists at all, we can register a callback for a graphql hook, which won't fire if the plugin isn't installed and activated:
```suggestion
		add_action( 'graphql_register_types', [ $this, 'graphql_register_types' ] );
```

kevinfodness (approved):
A few minor things to fix up before merge, otherwise 🍣

src/features/class-graphql.php
ellm: Yes! I left this by mistake after doing some testing locally. Thanks for flagging.

ellm: Thanks for flagging - this was left in by mistake after doing some local testing.

src/features/class-graphql.php
ellm: Ah yes, makes sense.

src/features/class-graphql.php
ellm: Updated in d6644c4e6727953425c9ad0899f4de0f041f1806

ellm:
Thanks @kevinfodness ! Any chance you know how best to fix the phpstan notices? Looks like it is flagging the WPGraphQL functions b/c there are no references to them in the codebase

kevinfodness:
@ellm The "right way" to fix those notices would be to install the WPGraphQL plugin as part of CI and add scandirs config to PHPStan to read in that content, but it would also require installing WPGraphQL as essentially a peer dependency when developing locally, and for what we're doing here, I'm not sure that's worth it. You have my blessing to `@phpstan-ignore` the parts that are complaining about not understanding the WPGraphQL functionality.

kevinfodness (approved):
🍣

ellm:
Thanks @kevinfodness - I also added some initial documentation on this in the Wiki - https://github.com/alleyinteractive/wp-curate/wiki/WPGraphQL-Integration

Issue #93, opened 2023-11-01
Prevent the current post from appearing in a curated list


mogmarsh:
### Description

If a WP Curate Query block is added to a post, and the rules are such that it can backfill with posts that would include the current post (e.g., the current post has a category of Entertainment and there is a Query block that looks for posts with a category of Entertainment) the post the user is on will show up in the list of autofilled results, and it should not.

### Use Case

A user should be able to use a Query block on a post and not have the post it's on show up in the recirc area.

### Acceptance Criteria

- Always exclude the current post from showing up in any auto-generated lists of posts when using the Query block.
- Should take effect on the front-end and in the Gutenberg backfill preview.

PR #162, opened 2024-04-04
Ensure the current post does not appear in auto generated lists of backfilled posts 

attackant:
Addresses #93 

Ensure the current post is not included in the any auto generated lists of backfilled posts (but the user can still choose it.) Also includes `npm audit fix` and a minor refactor for readability (hopefully)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to `WP Curate` will be documented in this file.
 
+## 1.8.3 - 2024-04-04
+
+- Enhancement: Exclude current post in backfilled posts query.
+
 ## 1.8.2 - 2024-03-27
 
 - Bug Fix: Error in 1.8.1 when Parse.ly is not instantiated.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 All notable changes to `WP Curate` will be documented in this file.
 
-## 1.8.3 - 2024-04-04
+## 1.9.0 - 2024-04-19
 
 - Enhancement: Exclude current post in backfilled posts query.
 
--- a/blocks/query/edit.tsx
+++ b/blocks/query/edit.tsx
@@ -80,7 +80,7 @@ export default function Edit({
       allowedTaxonomies = [],
       parselyAvailable = 'false',
     } = {},
-  } = (window as any as Window) || {};
+  } = (window as any as Window);
 
   const andOrOptions = [
     {
@@ -95,7 +95,7 @@ export default function Edit({
 
   // @ts-ignore
   const [isPostDeduplicating, postTypeObject] = useSelect(
-    (select: any) => {
+    (select) => {
       // @ts-ignore
       const editor = select('core/editor');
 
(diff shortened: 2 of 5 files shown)


mogmarsh:
I don't think this addresses the front end

mogmarsh:
I think to handle the front end, we should add a query class that will exclude the current post, then add that query class in https://github.com/alleyinteractive/wp-curate/blob/develop/src/features/class-query-block-context.php
I can try to help, or @dlh01 really knows this setup.

dlh01:
The `Exclude_Queries` class in Type Extensions could be used, if you can supply the current post ID to it: https://github.com/alleyinteractive/wp-type-extensions/blob/a67070fc2d03422c5e24f83a211bffea7845ffa3/src/alley/wp/post-queries/class-exclude-queries.php

attackant:
> The `Exclude_Queries` class in Type Extensions could be used, if you can supply the current post ID to it: https://github.com/alleyinteractive/wp-type-extensions/blob/a67070fc2d03422c5e24f83a211bffea7845ffa3/src/alley/wp/post-queries/class-exclude-queries.php

Thanks! I will check it out, I was just trying to roll my own

blocks/query/edit.tsx
```diff
@@ -169,21 +170,31 @@ export default function Edit({
...
+      apiFetch({ path }).then((response:any) => {
+        let revisedResponse;
+        if (Array.isArray(response)) {
+          revisedResponse = response.filter((item) => item !== currentPostId);
+        } else if (response.id === currentPostId) {
+          // If the current post is in the response, remove it.
+          revisedResponse = null;
+        } else {
+          revisedResponse = response;
+        }
```
mogmarsh: the response should always be an array. I'm not sure how to convince typescript of that.

mogmarsh (approved):
I think this looks good. As far as I can tell the logic should be the same on the front end. Did you get the same results in the editor and the front end with this change? And did you get the same results on the front end with the current production and this PR (with the exception of removing the current post)?
I'm going to create an issue to add some tests so we can make changes like this with more confidence.

src/features/class-query-block-context.php
```diff
@@ -20,99 +20,127 @@
...
+use Alley\WP\WP_Curate\Trending_Post_Queries;
 use WP_Block;
 use WP_Block_Type;
 use WP_Block_Type_Registry;
 
 /**
  * Provides context to query blocks
  */
-final class Query_Block_Context implements Feature {
+final readonly class Query_Block_Context implements Feature {
```
dlh01: As the CI results already noted, readonly classes aren't available while the plugin supports PHP 8.1.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to `WP Curate` will be documented in this file.
 
+## 1.8.3 - 2024-04-04
+
+- Enhancement: Exclude current post in backfilled posts query.
```
dlh01, line 7: Since there are no bugfixes in the release, it seems more like a minor release than a patch release to me, i.e. `1.9.0`.

dlh01 (commented):
Updates on the queries side look good to me!

PR #182, opened 2024-05-20
Add filters for query args

dlh01:
Adds filters for the query arguments used in the frontend and backend. The `current_post_id` parameter makes it possible to determine when the post being edited is also used as the site's front page.


scottnelle (approved):
👻

PR #193, opened 2024-06-19
Make it possible to clear the history of used post IDs

dlh01:
Swaps the `Used_Post_IDs` class for a custom implementation of `Post_IDs` that includes the ability to clear the history by firing an action. Includes a major version bump because of the changed constructor signatures.

### DIFF
--- a/src/features/class-query-block-context.php
+++ b/src/features/class-query-block-context.php
@@ -56,13 +56,11 @@ final class Query_Block_Context implements Feature {
 	}
 
 	/**
-	 * Filters the context provided to a query block
-	 * to determine the definitive list of backfilled posts.
+	 * Filters the context provided to a query block to determine the definitive list of backfilled posts.
 	 *
 	 * @param array<string, mixed>                 $context Default context.
 	 * @param array{"attrs": array<string, mixed>} $parsed_block Block being rendered.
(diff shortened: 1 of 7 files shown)


src/features/class-query-block-context.php
```diff
@@ -10,15 +10,14 @@
 use Alley\Validator\Comparison;
 use Alley\WP\Blocks\Parsed_Block;
 use Alley\WP\Post_IDs\Post_IDs_Envelope;
-use Alley\WP\Post_IDs\Used_Post_IDs;
 use Alley\WP\Post_Queries\Exclude_Queries;
 use Alley\WP\Post_Queries\Variable_Post_Queries;
 use Alley\WP\Types\Feature;
 use Alley\WP\Types\Post_Queries;
 use Alley\WP\Types\Post_Query;
-use Alley\WP\WP_Curate\Features\Parsely_Support;
```
mogmarsh, line 19: we're still referencing this... should it still be included?

mogmarsh (approved):
one question, otherwise 👍

src/features/class-query-block-context.php
mogmarsh: oh nevermind. same namespace

PR #201, opened 2024-07-15
Issue-200: Add a new block that can be nested or ability for Query Block to be nested

### LINKED ISSUE #200: Add a new block that can be nested or ability for Query Block to be nested
### Description

I would like a block that can be nested inside a Query block that has its own settings and can operate separately from the outer block.

### Use Case

On a homepage, the top story may have several optional related stories, such as in this screenshot from NYP.

Screenshot 2024-07-15 at 8 52 10 AM

The number of stories shown here, should be set without modifying the layout of the outer block and without modifying where pinned posts appear.

```
|----------------------------------------------------------|
|                                                          |
|           Pinned Story in Outer Block                    |
|                                                          |
|----------------------------------------------------------|
||---------------------------------------------------------|
||                                                        ||
||          New Inner Block With Independent Settings     ||
||          Could contain variable number of              ||
||          backfilled or pinned posts                    ||
||---------------------------------------------------------|
|----------------------------------------------------------|
|----------------------------------------------------------|
|                                                          |
|           Pinned Story in Outer Block                    |
|                                                          |
|----------------------------------------------------------|
|----------------------------------------------------------|
|                                                          |
|           Backfilled Story in Outer Block                |
|                                                          |
|----------------------------------------------------------|
```
For example, changing the number of posts shown in the inner block, would not affect the placement and which story is shown after it.

mogmarsh:
### Summary

This pull request introduces a new feature addressing the need for a block that can be nested inside a Query block with its own independent settings. This enhancement aims to provide flexibility in content layout, especially for use cases like homepage story arrangements where related stories need to be dynamically adjusted without affecting the main Query block's layout. Fixes #200

### Description

The addition of this new block allows for a nested structure within a Query block, enabling separate settings and operations from the outer block. This functionality is particularly useful for displaying a variable number of related stories or backfilled posts in a distinct section that does not influence the layout or settings of the surrounding content. During the development process, efforts were made to ensure that the subquery block could be added and configured with sidebar controls for customization. Debugging was performed to ensure the stability and functionality of these features.

### Use Case

A practical application of this feature can be seen in scenarios where a homepage features a top story with several optional related stories. The layout allows for these stories to be independently adjusted in number, without altering the main story's presentation or the arrangement of pinned posts within the outer Query block. This capability ensures greater flexibility and control over content display, catering to various layout needs and editorial decisions. The development process included iterations to refine the user experience in the editor and front-end rendering, ensuring that the subquery block behaves as expected, especially regarding initial rendering and the saving of inner blocks.

For further details, refer to the issue linked below:
[Issue #200](https://github.com/alleyinteractive/wp-curate/issues/200)


blocks/subquery/edit.tsx
```diff
@@ -0,0 +1,273 @@
+/* eslint-disable camelcase */
+import { useDebounce } from '@uidotdev/usehooks';
```
kevinfodness, line 4: Gutenberg has a useDebounce hook and would not require an additional dependency: https://github.com/WordPress/gutenberg/tree/a3d1c078cf88ff84a1d902cd86337dfa83c32642/packages/compose#usedebounce

blocks/subquery/edit.tsx
```diff
@@ -0,0 +1,273 @@
...
+import apiFetch from '@wordpress/api-fetch';
+import { InnerBlocks, useBlockProps } from '@wordpress/block-editor';
+import { useSelect } from '@wordpress/data';
+import {
+  useEffect,
+} from '@wordpress/element';
```
kevinfodness: WordPress has moved away from importing from `@wordpress/element` to importing directly from React, and we should do the same

blocks/subquery/edit.tsx
```diff
@@ -0,0 +1,273 @@
...
+      // @ts-ignore
+      const type = editor.getEditedPostAttribute('type');
+      // @ts-ignore
+      const meta = editor.getEditedPostAttribute('meta');
+
+      return [
+        // It's possible for usePostMetaValue() to run here before useEntityProp() is available.
+        Boolean(meta?.wp_curate_deduplication),
+        // @ts-ignore
+        type ? select('core').getPostType(type) : null,
```
kevinfodness, line 110: imo we should move anything with a `@ts-ignore` into the body of the function and ensure that what we're returning is of a stable type. The line above forces the first parameter to be a boolean, but here, it could be a post type object, null, or whatever `.getPostType` returns in a failure case (maybe undefined?) - I'd recommend moving the logic that gets the post type object before the `return`, validating that it is in fact a post type object, and only including it in the return value if it is actually a post type object, else return null. That way, even though this function is riddled with `@ts-ignore` because the Gutenberg project doesn't export types for `useSelect` that can be understood here, we can at least guarantee that what comes _out_ of this function is either a post type object or null for the second array member.

package.json
```diff
@@ -38,6 +38,8 @@
   },
   "dependencies": {
     "@alleyinteractive/block-editor-tools": "^0.10.2",
+    "@lukeed/uuid": "^2.0.1",
```
kevinfodness: Why this and not https://www.npmjs.com/package/uuid?

package.json
```diff
@@ -38,6 +38,8 @@
...
+    "@simbathesailor/use-what-changed": "^2.0.0",
```
kevinfodness: It looks like this is used for debugging - can we remove it before shipping an update to the plugin?

components/QueryControls/index.tsx
```diff
@@ -0,0 +1,274 @@
...
+  RangeControl,
+  SelectControl,
+  TextControl,
+  ToggleControl,
+} from '@wordpress/components';
+import { InspectorControls } from '@wordpress/block-editor';
+import {
+  Fragment,
+  createInterpolateElement,
+} from '@wordpress/element';
```
kevinfodness: Same here and elsewhere, let's import from React

components/QueryControls/index.tsx
```diff
@@ -0,0 +1,274 @@
...
+        >
+          <Checkboxes
+            label={__('Post Types', 'wp-curate')}
+            value={postTypes}
+            onChange={(newValue) => setAttributes({ postTypes: newValue, backfillPosts: [] })}
+            options={displayTypes}
+          />
+          {allowedTaxonomies.map((taxonomy) => (
+            <Fragment key={taxonomy.slug}>
+              { /* @ts-ignore */ }
```
kevinfodness, line 193: What is TS complaining about here?

components/QueryControls/index.tsx
```diff
@@ -0,0 +1,274 @@
...
+              />
+              {terms[taxonomy.slug]?.length > 1 ? (
+                <SelectControl
+                  label={sprintf(
+                    __('%s Relation', 'wp-curate'),
+                    taxonomy.name,
+                  )}
+                  help={__('AND: Posts must have all selected terms. OR: Posts may have one or more selected terms.', 'wp-curate')}
+                  options={andOrOptions}
+                  onChange={(newValue) => setTermRelation(taxonomy.slug, newValue)}
```
kevinfodness: We generally standardize on `next` as the new value, and there's a mix of both in this file.

services/deduplicate/index.ts
```diff
@@ -118,6 +119,22 @@ export function mainDedupe() {
     });
   }
 
+  if (queryBlocks.length !== allQueryBlocks.length) {
+    allQueryBlocks.forEach((queryBlock) => {
+      // if (queryBlock.name === 'wp-curate/query' && queryBlock.attributes.query) {
```
kevinfodness: Remove

kevinfodness (approved):
Left some suggestions and questions, I'll leave it up to you what you implement vs. not 🍣

components/QueryControls/index.tsx
mogmarsh: @kevinfodness Term Selector has [default props](https://github.com/alleyinteractive/alley-scripts/blob/main/packages/block-editor-tools/src/components/term-selector/index.jsx#L42-L50), but somehow we're not communicating that properly and TS wants them included.

blocks/subquery/edit.tsx
mogmarsh: I decided to keep the existing one, since it operates on the value. The Gutenberg one wants to debounce a function. We can look at refactoring this later if we want.

Issue #48, opened 2023-09-12
Option for pinned posts to be unique no matter where they occur on the page


mogmarsh:
### Description

Add a filterable setting to where a pinned post anywhere on the page will not be shown earlier on the page by backfill, so that a post only shows up once. Several clients have asked for this model.
This setting needs to be read and handled in the admin and on the front end.
There is commented out code in the admin. We have prior art for the front end.

### Use Case

Currently it is set up so that if a pinned post is backfilled on the page before it is pinned, then it will show twice. Several clients have asked for the posts to be totally unique on the page, so we should allow for this behavior.

### Acceptance Criteria

- Modify the behavior of deduplication such that it excludes all pinned posts in all blocks on a view, regardless of where they appear.
- Ensure that if a post would appear in an automatically filled area at the top of the page, but is pinned further down on the page, it does not appear twice (should be deduplicated from the automatically filled section at the top).\
- Should take effect on the front-end and in the Gutenberg backfill preview.

efuller:
Closed by #202

PR #202, opened 2024-07-15
Feature: Unique Pinned Posts

efuller:
Closes ticket #48 

## Description
Provides functionality that will allow for unique, pinned posts on a page. This means that if a post is pinned, it will not display anywhere else on the page.

## Notes
- Uniqued pinned posts relies on deduplication being enabled (the editor toggle should reflect this)
- Will bump version after CR

## Additional Updates
- GH action workflows have been updated to allow for tests to be run properly
- Mantle Testkit has been updated


composer.json
```diff
@@ -54,7 +58,7 @@
   "prefer-stable": true,
   "scripts": {
     "phpcbf": "phpcbf .",
-    "phpcs": "phpcs .",
+    "phpcs": "phpcs . -sn",
```
kevinfodness: We shouldn't define these flags here. The `s` flag is already defined in `.phpcs.xml` and we don't want to suppress warnings.

entries/slotfills/deduplication/deduplication.tsx
```diff
@@ -2,23 +2,33 @@ import { ToggleControl } from '@wordpress/components';
...
 import { usePostMetaValue } from '@alleyinteractive/block-editor-tools';
-
-/**
- * The following code is a temporary fix.
- *
- * Once the issue linked below is resolved, this code can be removed.
- * @link https://github.com/alleyinteractive/alley-scripts/issues/473
- */
 import { useSelect } from '@wordpress/data';
+import { useEffect } from '@wordpress/element';
```
kevinfodness: We should import from React instead. The WP project has transitioned away from using `@wordpress/element` to using React directly.

services/deduplicate/index.ts
```diff
@@ -96,26 +96,28 @@ export function mainDedupe() {
...
-  // });
+
+  /**
+   * This block of code is responsible for enforcing the unique pinned posts setting in the editor.
+   */
+  if (wpCurateUniquePinnedPosts) {
+    queryBlocks.forEach((queryBlock) => {
+      const { attributes } = queryBlock;
+      const { posts } = attributes;
+      posts?.forEach((post) => {
```
kevinfodness: We can collapse this quite a bit:
```suggestion
    queryBlocks.forEach((queryBlock) => {
      queryBlock?.attributes?.posts?.forEach((post) => {
```

src/class-supported-post-types.php
```diff
@@ -92,6 +92,14 @@ public function register_post_meta(): void {
 				'type' => 'boolean',
 			]
 		);
+		register_meta_helper(
+			'post',
+			$this->get_supported_post_types(),
+			'wp_curate_unique_pinned_posts',
+			[
+				'type' => 'boolean',
```
kevinfodness: We should set a default of `false`

src/post-ids/class-pinned-in-post-content.php
```diff
@@ -0,0 +1,75 @@
...
+		if ( $main_query->is_singular() ) {
+			$post = $main_query->get_queried_object();
+
+			if ( $post instanceof \WP_Post ) {
+				// Unique pinned posts relies on deduplication being enabled.
+				$post_level_unique         = get_post_meta( $post->ID, 'wp_curate_unique_pinned_posts', true );
+				$post_level_deduplication  = get_post_meta( $post->ID, 'wp_curate_deduplication', true );
+				$post_level_unique_enabled = $post_level_unique && $post_level_deduplication;
+
+				if ( true === (bool) $post_level_unique_enabled ) {
```
kevinfodness: No need to do several casts to boolean here, the `&&` will do it:
```suggestion
				if ( $post_level_unique && $post_level_deduplication ) {
```

tests/feature/UniquePinnedPostTest.php
```diff
@@ -0,0 +1,148 @@
...
+	 * Run before each test.
+	 */
+	protected function setUp(): void {
+		parent::setUp();
+
+		$this->set_permalink_structure( '/%postname%/' );
+
+		static::factory()->post->create(
+			[
+				'post_title' => 'TheWrapTestPost 1',
```
kevinfodness: We shouldn't have client names in tests in open source repos. Let's just make this "Test Post 1" etc for others in this file.

kevinfodness (changes requested):

efuller:
@kevinfodness I've adjusted code per your review and requested another review. Could you have a look when you have a moment?

kevinfodness (approved):
🍣

PR #205, opened 2024-07-17
Adds filter to control max number of posts in query block

### LINKED ISSUE #114: Make Max Number of Posts Filterable
### Description

The maximum number of posts is hardcoded to 10, but should be filterable. It's used in Gutenberg here to set the upper bound of the range control: https://github.com/alleyinteractive/wp-curate/blob/develop/blocks/query/edit.tsx#L293

### Use Case

> As a developer, I want to be able to modify the max number of posts that is available in the range control in a Query block.

### Acceptance Criteria

- Add a filter in PHP to set the max number of posts for the query block. Default it to 10.
- Use `wp_localize_script` to make this value available to the editor script for the block.
- If this value is set, have the edit component for the block use it instead of the default that comes in from attributes.

willowCeleste:
This adds a filter that allows a dev to control the max number of posts allowed in the query block. I removed the old hardcoded attribute `maxNumberOfPosts` and replaced it with a new value which is passed into the block via `wp_localize_script`.

Fixes #114


mboynes (approved):
Code is 💯.

One outstanding question on this though, that I can't answer but maybe @mogmarsh can: by removing the attribute we eliminate the opportunity to add a UI control for this value. Is that OK? Is that something that would be valuable?

blocks/query/index.php
```diff
@@ -43,6 +43,13 @@ function wp_curate_query_block_init(): void {
 	 */
 	$allowed_taxonomies = apply_filters( 'wp_curate_allowed_taxonomies', [ 'category', 'post_tag' ] );
 
+	/**
+	 * Filter the taxonomies that can be used in the Query block.
```
mogmarsh:
```suggestion
	 * Filter the maximum number of posts that can be displayed in the Query block.
```

mogmarsh:
@mboynes @willowCeleste 
The original thought behind this was that a developer could set the `maxNumberOfPosts` and `minNumberOfPosts` in code - possibly in a template or a block theme, etc. and the end user would not see the attribute (unless they went into code view), but would be restricted in how many posts they could select on the front end. 

For instance, the developer could say "you can have between 3 and 5 posts in this particular location", which gives the end user some flexibility, but not full control. If the minimum and maximum are set to the same number, the end user does not get any control, the range selector is not displayed.

I wonder if maybe we can leave support for `maxNumberOfPosts` and `minNumberOfPosts` but instead of setting the default value of `maxNumberOfPosts` to `10` we use the new localized value? I think that would allow for the original functionality, but also allow for a sitewide max.

kevinfodness:
imo I'd rather not have the "invisible setting" and go with a sitewide max, and allow flexibility in use of the blocks.

blocks/query/index.php
```diff
@@ -43,6 +43,13 @@ function wp_curate_query_block_init(): void {
 	 */
 	$allowed_taxonomies = apply_filters( 'wp_curate_allowed_taxonomies', [ 'category', 'post_tag' ] );
 
+	/**
+	 * Filter the maximum number of posts that can be displayed in the Query block.
+	 *
+	 * @param integer $max_posts The maximum number of posts to display.
+	 */
+	$max_posts = apply_filters( 'wp_curate_max_posts', 5 );
```
kevinfodness: This should default to the previous default, which is 10:
```suggestion
	$max_posts = apply_filters( 'wp_curate_max_posts', 10 );
```

PR #206, opened 2024-07-18
Issue-203: Drag and Drop posts within editor

### LINKED ISSUE #203: Drag and Drop posts within editor
### Description

Either using the existing Gutenberg drag and drop controls, or by adding a new drag handle to the Post block, we should add functionality to move a post to a new spot in the editor.

Each Post block should have a drag handle or use the Gutenberg one if possible.
Each Post block should be a drop zone.
The drag action should include a payload with the Post ID, any Post customizations, and the block ID of the Post being dragged.

If a drop action is received, the Post info in the payload should be set on the receiving block, and the pinned post should be removed from the originating block via the block ID.

Drag and drop should not affect the layout of the block or any block attributes such as class names, etc. The post should be moved to the new location, and that is all.

If a post was previously pinned in the receiving block, it should just be wiped out (or maybe swapped). We don't want to shift posts. 

For first implementation, let's just worry about dragging and dropping within the same Query block, but if it happens to work across Query blocks, that's great.

### Use Case

Moving a pinned post from one slot to another should be easier.

mogmarsh:
### Summary

Implements drag and drop functionality for posts within the editor to make moving a pinned post from one slot to another easier. Fixes #203.

### Test Plan

- [ ] Ensure each Post block has a drag handle or uses the Gutenberg one if possible.
- [ ] Confirm each Post block acts as a drop zone.
- [ ] Test the drag action includes a payload with the Post ID, any Post customizations, and the block ID of the Post being dragged.
- [ ] Verify on drop, the Post info in the payload is set on the receiving block, and the pinned post is removed from the originating block via the block ID.
- [ ] Drag and drop should not affect the layout or any block attributes such as class names.
- [ ] Check if a post was previously pinned in the receiving block, it is properly handled.
- [ ] Improved the cancel and move process for a smoother drag and drop experience.
- [ ] Fixed an issue where moving a post did not work as expected.
- [ ] Removed the click handler to ensure that drag and drop functionality is streamlined and intuitive.

### Additional Notes

- Styling has been updated to include a cursor pointer on hover to indicate draggable elements.
- Efforts were made to reduce border thickness and fix outline styles for a cleaner UI appearance during drag and drop operations.
- Code has been linted and reworked for better performance and readability.


mboynes (approved):
💯. Reviewed and tested, works great!

PR #208, opened 2024-07-19
Add custom Post Title block

### LINKED ISSUE #204: Allow title/headline overrides
### Description

Create a custom block that extends or replaces the [core Post Title block](https://wordpress.org/documentation/article/post-title-block/)
The block should allow for a custom headline to be set for the post, that will be stored in the attributes of the Query block.
If a custom headline is set, it will be used in place of the original headline in the editor, on the front end, and in graphql. Deleting the custom headline will revert back to the original headline and remove the value from the Query block attributes.
The core block allows for changing the Post Title, but this change is saved back to the post as the new headline, which is not the behavior we want. This override should only apply to the current Query block.

This functionality only needs to be available to pinned posts. If a post is no longer pinned in the Query block, the title override back be removed from the attributes.

Update the Query block inner blocks template to use the new block.

### Use Case

Editors should be able to set an override title for use in the Query block.

ellm:
- This PR adds a new `wp-curate/post-title` block and replaces the `core/post-title` block use in the post template. 
- The new post title block allows for a custom title to be used within a Query block instance. The custom title will only exist within the context of the Query block and does not change the post title associated with the post object.
- This work supports core WordPress editorial and frontend. Follow-up work will happen in a separate PR to support WPGraphQL integration.


blocks/post-title/render.php
```diff
@@ -0,0 +1,46 @@
...
+
+if ( empty( $post_title ) || empty( $post_link ) ) {
+	return;
+}
+
+// Use custom post title, if available.
+if ( 0 < count( $custom_post_titles ) ) {
+	foreach ( $custom_post_titles as $value ) {
+		if ( $value['postId'] === $current_post_id ) {
+			$post_title = $value['title'];
```
mogmarsh, line 36:
```suggestion
			$post_title = $value['title'];
			break;
```

mogmarsh (approved):
this looks good 👍

PR #217, opened 2024-08-15
Issue-216: Pinned posts that are deleted or unpublished cause problems in editor and on graphql front end

### LINKED ISSUE #216: Pinned posts that are deleted or unpublished cause problems in editor and on graphql front end
### Description of the bug

If a pinned post is deleted or unpublished, it causes issues in the editor (you can't tell that the post has been pinned) and on the front end in graphql.

### Steps To Reproduce

Pin a post
Unpublish that post

### Additional Information

_No response_

mogmarsh:
### Summary

Fixes #216 - This pull request addresses the issue where pinned posts that are deleted or unpublished cause problems in the editor and on the graphql front end.

### Description

If a pinned post is deleted or unpublished, it causes issues in the editor (you can't tell that the post has been pinned) and on the front end in graphql. The solution now includes showing an error in the editor when a pinned post is deleted or unpublished and removing the post's ID from the graphql front end to prevent errors.

### Steps To Reproduce

1. Pin a post
2. Unpublish that post

### Additional Information

_No response_


services/deduplicate/index.ts
```diff
@@ -118,7 +121,7 @@ export function mainDedupe() {
   }
 
   // Loop through all query blocks and set backfilled posts in the open slots.
-  queryBlocks.forEach((queryBlock) => {
+  queryBlocks.forEach(async (queryBlock) => {
```
efuller, line 124: While this code does look to work, eventually we will want to refactor the usage of an async callback here. Some recommendations call for use of a `for...of` loop but it can largely depend on what other functionality we have going on here.

efuller (approved):
⚡ For transparency, I did make one change after checking with Greg. Also noted one of item for future reference.

PR #219, opened 2024-08-16
Feature: Issue-218: Cache Initial Post API Calls

### LINKED ISSUE #218: Cache Initial API Calls for Query Blocks
# Summary
When loading a page that is powered by WP-Curate, for each query block, there is an API request that is made. We would like to find a way to optimize these API calls so that each block doesn't have to make the same calls.

# Description

# Use Case

efuller:
## Summary
This PR introduces `SWR` which will help to cache API responses that are identical calls. The `useSWRImmutable` hook is used so that automatic revalidation is not turned on.

## Reviewer Notes
Will bump version after CR.

## Description
When loading a page in the editor that is powered by the query block, there are several API calls to the API for gathering the posts needed. Many of these API calls are the same. Find a way to increase performance by minimizing the number of API calls upon load.

## Testing
Before merging in, it would be great to get another set of 👀 on actual testing in the editor. 
1. Verify that query changes are reflected in real time in the editor
2. Verify that deduplication is working
3. Verify that pinned posts are working as they should

## Ticket
Fixes #218


kevinfodness (approved):
🍣

blocks/query/edit.tsx
```diff
@@ -1,4 +1,5 @@
 /* eslint-disable camelcase */
+import useSWRImmutable from 'swr/immutable';
```
dlh01, line 2: @efuller Does disabling automatic revalidation eliminate the need for workarounds like this one? https://l.alley.dev/ca8db44a8f

efuller: @dlh01 I _think it could_. `useSWRImmutable` will simply just cache an API call and unless that request changes in any way, it will continue to use the cache. 

The use case I wanted to guard against was the revalidation after switching tabs. [Here are some docs](https://swr.vercel.app/docs/revalidation#disable-automatic-revalidations) on revalidation and there are a ton of [additional options](https://swr.vercel.app/docs/api#options) as well.

I think it could be worth a try!

PR #221, opened 2024-08-22
Enhancement - Add Filter for Post Title Heading Level Support

ellm:
# Summary

Provides ability within projects to disable the Heading Level dropdown from the Post Title block settings. 

Example

```
wp.hooks.addFilter( 'blocks.registerBlockType', 'wp-curate/example', ( settings, name ) => {
    const maybeFilteredSettings = { ...settings };

    if( 'wp-curate/post-title' === name ) {
        maybeFilteredSettings.attributes.supportsLevel.default = false;
    }

    return maybeFilteredSettings
});
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,13 +2,17 @@
 
 All notable changes to `WP Curate` will be documented in this file.
 
+## 2.2.3 - 2024-08-22
+
+- Enhancement: Enable block filter support for Post Title block setting Heading Level select.
+
 ## 2.2.2 - 2024-08-21
 
 - Enhancement: Introduce `SWR` for caching API requests in the Query block.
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -2,13 +2,17 @@
 
 All notable changes to `WP Curate` will be documented in this file.
 
+## 2.2.3 - 2024-08-22
```
goodguyry: I'd argue this is more than a patch and could/should be `2.3.0`

goodguyry (approved):
🎸

CHANGELOG.md
ellm: Yeah, I suppose that is accurate. I've made that update. Thanks for flagging

PR #225, opened 2024-08-26
make buttons full width at small sizes

mogmarsh:


blocks/post/index.scss
```diff
@@ -56,3 +58,21 @@
...
+  .wp-curate-post-block__post-picker {
+    width: 100%;
+  }
+
+  .wp-curate-post-block__actions, .components-button-group {
+    display: flex;
+    flex-direction: column;
+    width: 100%;
+
+      button {
```
manewc: looks like the tabbing is off here

manewc (approved):
🍿

PR #226, opened 2024-08-26
Hotfix/fix async dedupe

mogmarsh:
Move the manual post validation from mainDedupe() to a useEffect() in the query block edit. This prevents the posts from getting backfilled out of order.


manewc (approved):
🥼

PR #229, opened 2024-09-04
Take 2 of "Adds filter to control max number of posts in query block"

mogmarsh:
Allows for the `maxNumberOfPosts` attribute to be set on a block, but also allows for the sitewide maxPosts to be set in php. The block will use whichever value is set and is lower.


attackant (approved):
🐙

ellm:
Flagging https://github.com/alleyinteractive/wp-curate/issues/241 here for consideration on this PR

Issue #230, opened 2024-09-09
Editor crashes when using transforms menu


samhermes:
### Description of the bug

When using the Query block, if a user opens the transforms menu, the block will crash when they hover over one of the options.

### Steps To Reproduce

1. Insert a Query block
2. Select the Query block
3. Click on the Query block icon in the block toolbar to open the transforms menu
4. Hover over one of the options (defaults to Columns and Group blocks)
5. The editor crashes

### Additional Information

The deduplication logic is running when a user opens the transforms menu, as the transforms menu uses a copy of the current block to display what the transformed block will look like.

PR #231, opened 2024-09-09
Skip deduplication during block transform

samhermes:
Addresses #230 by preventing deduplication from running when the transforms preview element is present.


mogmarsh (approved):
👍

mogmarsh:
@samhermes If you need this in a release right away so you can test it in the client site, then you'll need to bump the version number and add to the changelog. Merging it with a bumped version number will cause a new release.

PR #234, opened 2024-09-16
Hotfix/unpublished posts filter

mogmarsh:
use `array_map` instead of `array_filter` so we hold the space for curated posts that were deleted

### DIFF
--- a/src/class-plugin-curated-posts.php
+++ b/src/class-plugin-curated-posts.php
@@ -74,9 +74,9 @@ final class Plugin_Curated_Posts implements Curated_Posts {
 		}
 
 		$pinned_posts = $attributes['posts'] ?? $block_type->attributes['posts']['default'];
-		$pinned_posts = array_filter( $pinned_posts, function ( $id ) {
-			return 'publish' === get_post_status( $id );
-		} );
+		$pinned_posts = array_map( function ( $id ) {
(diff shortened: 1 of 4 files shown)


src/class-plugin-curated-posts.php
```diff
@@ -74,9 +74,12 @@ public function with_query_context( array $context, array $attributes, WP_Block_
...
 		$pinned_posts = $attributes['posts'] ?? $block_type->attributes['posts']['default'];
-		$pinned_posts = array_filter( $pinned_posts, function ( $id ) {
-			return 'publish' === get_post_status( $id );
-		} );
+		$pinned_posts = array_map( function ( $id ) {
+			if ( ! $id ) {
+				return null;
+			}
+			return 'publish' === get_post_status( $id ) ? $id : null;
+		}, $pinned_posts );
```
mboynes: logic can be simplified to `$id && 'publish' === get_post_status( $id )`... which means this could use an arrow function if you felt so inclined:

```suggestion
		$pinned_posts = array_map( fn ( $id ) => $id && 'publish' === get_post_status( $id ) ? $id : null, $pinned_posts );
```

mboynes (approved):
🚀 an optional suggestion for you, but it's also just dandy as-is

PR #235, opened 2024-09-17
Enhancement - Enable support for hiding post type selection from block settings UI

ellm:
Provides a method for hiding specific post types from the Query Parameters section of the block settings UI. 

Since the configuration comes from block attributes on the Query block, it allows for more control at the individual block level when globally supported post types should not be included.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,14 @@
 
 All notable changes to `WP Curate` will be documented in this file.
 
+## 2.5.0 - 2024-09-17
+
+- Enhancement: Enable support for hiding post type selection from block settings UI.
+
 ## 2.4.0 - 2024-08-28
 
 - Enhancement: Subquery block added to allow a separate set of posts within a query block.
-see https://github.com/alleyinteractive/wp-curate/issues/200
+see <https://github.com/alleyinteractive/wp-curate/issues/200>
 
 ## 2.3.3 - 2024-09-16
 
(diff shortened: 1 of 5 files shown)


wp-curate.php
```diff
@@ -3,7 +3,7 @@
  * Plugin Name: WP Curate
  * Plugin URI: https://github.com/alleyinteractive/wp-curate
  * Description: Plugin to curate homepages and other landing pages
- * Version: 2.4.0
+ * Version: 2.5.0
```
mogmarsh: I feel like this could be 2.4.1. Installing it won't break anything.

CHANGELOG.md
```diff
@@ -2,10 +2,14 @@
 
 All notable changes to `WP Curate` will be documented in this file.
 
+## 2.5.0 - 2024-09-17
```
mogmarsh: I feel like this could be 2.4.1. Installing it won't break anything.

mogmarsh (approved):
👍

wp-curate.php
ellm: Updated in [0140c29](https://github.com/alleyinteractive/wp-curate/pull/235/commits/0140c296932d93cf1309a469b72f499ae5ff8b89)

CHANGELOG.md
ellm: Updated in [0140c29](https://github.com/alleyinteractive/wp-curate/pull/235/commits/0140c296932d93cf1309a469b72f499ae5ff8b89)

PR #238, opened 2024-09-27
Issue-237: `updateValidPosts` check does not pass post types and only checks for valid posts, other post types fail

### LINKED ISSUE #237: `updateValidPosts` check does not pass post types and only checks for valid posts, other post types fail
### Description of the bug

When pinning a post from a custom post type (not post), you receive a warning that the post has been deleted and you cannot pin the post.

### Steps To Reproduce

Pin a post not of type 'post'

### Additional Information

_No response_

mogmarsh:
## Summary
This PR addresses the issue described in https://github.com/alleyinteractive/wp-curate/issues/237 - Fixes #237

### Description of the bug

When pinning a post from a custom post type (not post), you receive a warning that the post has been deleted and you cannot pin the post.

### Solution

The validation has been updated to allow additional post types, ensuring that posts of custom types can be pinned without triggering a deletion warning.

### Steps To Reproduce

Pin a post not of type 'post'

### Additional Information

_No response_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to `WP Curate` will be documented in this file.
 
+## 2.4.2 - 2024-09-27
+
+- Bug Fix: Posts not of type 'post' are flagged as being deleted.
+
 ## 2.4.1 - 2024-09-19
 
 - Enhancement: Enable support for hiding post type selection from block settings UI.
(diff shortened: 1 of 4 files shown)


mboynes (approved):
🚀 Looks great! Would be nice to have a unit test accompanying a fix like this, if that's feasible.

PR #279, opened 2025-01-13
Issue-278: In the title block, manually setting a custom title back to the original title does not work

### LINKED ISSUE #278: In the title block, manually setting a custom title back to the original title does not work
### Description of the bug

If you set a custom title for a pinned post, and then try to manually set the title back to the original title, it will not save - it will keep the last version that was different.
For example, if my post title is "My Post Title", and I set it to "My Post Title Custom", if I try to delete " Custom" from then end, it will set the title to "My Post Title C", because that was the last version that was different from "My Post Title"
I believe the bad line is https://github.com/alleyinteractive/wp-curate/blob/develop/blocks/post-title/edit.tsx#L96
If you delete the whole custom title that does reset the title back to the original. The code should do the same thing. If the custom post title equals the original post title, set the custom title to an empty string.

### Steps To Reproduce

1. Pin a post
2. Set a custom post title
3. Manually try to set the title back to the original title
4. Save and refresh and you will see the last version of the title that was different.

### Additional Information

_No response_

mogmarsh:
**Summary:**
This PR addresses the issue where manually setting a custom title back to the original title in the title block does not work as expected. Specifically, if a custom title is set for a pinned post and later attempts are made to revert it back to its original title, the changes do not save correctly. The problem seems to be related to the handling of custom titles in the code. Fixes #278

**Changes:**
- Modify the title handling logic to ensure that setting a title back to its original value is properly saved.
- Implemented a fix allowing users to manually reset the custom post title back to its original.

**Test Plan:**
1. Pin a post.
2. Set a custom post title.
3. Attempt to manually revert the title back to the original title.
4. Save and refresh to confirm that the title has been successfully reverted to its original form.

**Additional Information:**
- Issue link: [Issue #278](https://github.com/alleyinteractive/wp-curate/issues/278)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to `WP Curate` will be documented in this file.
 
+## 2.4.9 - 2025-01-13
+
+- Bug Fix: Allow manually resetting custom post title back to original title.
+
 ## 2.4.8 - 2024-12-17
 
 - Bug Fix: Roll back to React 18 again.
--- a/blocks/post-title/edit.tsx
+++ b/blocks/post-title/edit.tsx
@@ -92,10 +92,7 @@ export default function Edit({
      * Handle cases for when it's not necessary to update the collection of
      * custom titles
      */
-    if (
-      title === rawTitle
-      || title === currentCustomPostTitle?.title
-    ) {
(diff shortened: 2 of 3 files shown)


ellm (approved):
👍 Nice! I checked this out locally and tested too. It is working as described.

PR #283, opened 2025-02-05
Update Parsely integration for 3.17

dlh01:
Fixes #281.

PR #284, opened 2025-02-07
Revert "Update Parsely integration for 3.17"

dlh01:
Reverts alleyinteractive/wp-curate#283


dlh01:
Merging this revert because of the possibility of a fatal error and because, as a patch release, most installations will use a version range that will cause them to be auto-updated.

PR #332, opened 2025-07-10
Add filter for deduplication

mslinnea:
Adds filter to allow custom logic for deduplication.


mogmarsh (changes requested):
This looks good, but I have a couple questions/suggestions.
I wonder if we want to change the other returns in this section (lines 91, 98) into variable assignment and then have one filtered return of that variable at the end? This way the variable would filter all cases.
Either that or change the filter to `wp_curate_default_deduplication_enabled` or something like that, because this only overrides the default `false` value, not anything set in the block attributes or post meta. I wouldn't want someone to think they could filter all behavior and think it's not working right.
Also, I don't know what other changes you have planned, but the plugin will need a readme update and version bump so it builds a new version.

mslinnea:
> This looks good, but I have a couple questions/suggestions. I wonder if we want to change the other returns in this section (lines 91, 98) into variable assignment and then have one filtered return of that variable at the end? This way the variable would filter all cases. Either that or change the filter to `wp_curate_default_deduplication_enabled` or something like that, because this only overrides the default `false` value, not anything set in the block attributes or post meta. I wouldn't want someone to think they could filter all behavior and think it's not working right. Also, I don't know what other changes you have planned, but the plugin will need a readme update and version bump so it builds a new version.

I'm going to update this as we discussed and then merge it first to the `fam` branch so that we can test it on our project before doing a release.

PR #349, opened 2025-08-04
Issue-343: WordPress Version 6.8.2 Breaks Subquery Block

### LINKED ISSUE #342: Cannot Add Subquery Block
### Description of the bug

The changes in #333 seems to prevent the Subquery block from be added in the appropriate place. 

It needs to be added inside of the `wp-curate/post` block like so:
Image

Perhaps instead of specifying a parent of `wp-curate/query`, we specify the `wp-curate/post` as the required parent. 

### Steps To Reproduce

1. Add a Query block
2. Try to add a Subquery block after the excerpt. It shouldn't be available to choose.

### Additional Information

_No response_

### LINKED ISSUE #343: WordPress Version 6.8.2 Breaks Subquery Block
### Description of the bug

When using a Subquery block on WP 6.8.2, it causes the browser to freeze up. If I revert to WordPress 6.7.2, the Subquery block works as expected.

https://github.com/user-attachments/assets/5bb9e6c3-3570-478b-ae09-8a3a829fab95

### Steps To Reproduce

1. Make sure to have a WordPress install at version 6.8.2
2. Add a Query block
3. Try adding a Subquery block

### Additional Information

_No response_

efuller:
## Summary

Fixes #343  
Fixed #342

This pull request resolves an issue where using a Subquery block on WordPress 6.8.2 caused the browser to freeze. The root cause was the parent restriction on the Subquery block and logic in the block's edit component. The Subquery block now works as intended in WordPress 6.8.2.

## Developer Notes

- The allowed parent block for the Subquery block has been changed from `wp-curate/query` to `wp-curate/post` (see `blocks/subquery/block.json`). This ensures the Subquery block is only inserted in the correct context and resolves the freeze.
- The Subquery block's edit component has been refactored:
  - TypeScript types are now stricter, with the `posts` attribute typed as an array of `WpRestApiPost['id'] | null` rather than `any[]`.
  - Data fetching with `useSWRImmutable` is now handled safely and never called conditionally; instead, it’s passed `null` when not needed.
  - Initialization and normalization logic has been improved. State updates and deduplication are now handled in dedicated `useEffect` hooks.
  - The code avoids directly mutating the `posts` attribute and uses a normalized approach for state updates.
  - Several variables and expressions were clarified and extracted for readability.
- TypeScript issues in the code are now clearly marked with `@ts-ignore` and TODO comments for future resolution.

## How Has This Been Tested?

1. Set up a WordPress installation at version 6.8.2.
2. Add a Query block.
3. Add a Subquery block inside a Post block and verify that the browser does not freeze.

## Screenshots (if appropriate)

[User attachment reference](https://github.com/user-attachments/assets/5bb9e6c3-3570-478b-ae09-8a3a829fab95)

## Checklist

- [ ] My code follows the code style of this project.
- [ ] I have performed a self-review of my own code.
- [ ] I have added tests that prove my fix is effective or that my feature works.
- [ ] New and existing unit tests pass locally with my changes.
- [ ] I have updated the documentation accordingly.


blocks/query/types.ts
```diff
@@ -6,7 +8,7 @@ interface EditProps {
     minNumberOfPosts?: number;
     numberOfPosts?: number;
     offset?: number;
-    posts?: any[];
+    posts?: Array<WpRestApiPost['id'] | null>;
```
kevinfodness, line 11: Nice!

blocks/subquery/edit.tsx
```diff
@@ -142,12 +140,23 @@ export default function Edit({
...
+  useEffect(() => {
+    if (!isFirstPost) {
+      return;
+    }
+
+    if (postTypes?.length > 0) {
+      return;
+    }
+
+    setAttributes({ postTypes: allowedPostTypes.map((type) => type.slug) });
```
kevinfodness: Could use destructuring here:
```suggestion
    setAttributes({ postTypes: allowedPostTypes.map(({ slug }) => slug) });
```

blocks/subquery/edit.tsx
```diff
@@ -208,22 +223,48 @@ export default function Edit({
               context: 'edit',
             },
           ),
-        }).then((response) => (response as any as WpRestApiPosts).map((post) => post.id));
+        }).then((result) => {
+          const resultIds = result.map((post) => post.id);
+          setAttributes({ validPosts: resultIds });
+        });
```
kevinfodness: With `await` I wouldn't use `then`, I'd set the return value. I'd also not trust that the return from the API fetch is what we think it is. Something like:
```ts
const response = await apiFetch<unknown>({ /* etc */ });
const resultIds: number[] = Array.isArray(response)
  ? response.map((post: unknown) => Number.isInteger(post?.id) ? post.id : 0).filter()
  : [];
setAttributes({ validPosts: resultIds });
```

kevinfodness (changes requested):
The only bit of feedback I care strongly about is to not trust what comes back from the REST API. Yes, it _should_ return an array of post objects, but if it doesn't, the code won't work as expected. I like to lean in to TS here and explicitly say that the value is `unknown` and have TS make me prove that it's what I say it is (by checking first).

blocks/subquery/edit.tsx
efuller: @kevinfodness I've made some adjustments here, if you could take another look 🙏 

I am curious what the use case might be that we can't trust the core posts endpoint?

kevinfodness: We really shouldn't trust data coming from _anywhere_ outside of a TypeScript script, whether that's an API, user input, a database, etc. We know what it _should_ be, but we don't know what it _is_ without validation. The true power of TypeScript is when we can confidently pass data through our application knowing its shape, but if we accept data from an API endpoint and use it without validation, it's possible that different data is returned (because of an error, or because of an update to WordPress core or a plugin or custom code, etc). Validation doesn't "cost" much (either in terms of time to implement or in terms of computational time) and it increases the faith we have in our code working as expected, so it's worth the investment, imo.

blocks/subquery/edit.tsx
```diff
@@ -161,73 +169,117 @@ export default function Edit({
...
+            _locale: 'user',
+            context: 'edit',
+          },
+        ),
+      });
+
+      const resultIds = Array.isArray(result)
+        ? result
+          .filter(
+            (post): post is { id: number } => post
```
kevinfodness: Could use `unknown` here too, which should be fine because of the checking you're already doing, and it tells TypeScript to explicitly validate that the checks you are doing are correct for dealing with unknown data.

kevinfodness: Note that you may need to do this in a `map` instead of a `filter`, where you return `0` if you can't get down to an ID and then run the filter afterwards where you filter out the zeroes.

kevinfodness (approved):
One comment about the data validation, otherwise 🍣

blocks/subquery/edit.tsx
efuller: @kevinfodness how does 49aea00 look?

kevinfodness: Perfect!

blocks/subquery/edit.tsx
efuller: Makes sense! I think I was getting tripped up in my brain because it wasn't an 'outside' API, but that doesn't really matter. Ty for the explanation.

PR #362, opened 2025-08-29
Issue-158: Add \"Order By\" options

### LINKED ISSUE #158: Add "Order By" options
### Description

I would like to see `orderby` options included with the query block to allow more control over the order of posts that appear.

### Use Case

When an editor uses the query block, they should be able to define the order by parameter for backfilled posts. At a minimum, our block should allow the same ordering as the core query loop block:

Screenshot 2024-03-26 at 4 10 37 PM

### LINKED ISSUE #169: Allow for sorting by post field or meta value
### Description

It could be useful to sort posts in a query block by a post field meta value. The meta field options should be filterable by php - with a meta_key and a display name.
We should be able to support ascending or descending.

### Use Case

In the case of content that is not "posts", we may want to order by the title or by a meta field with a separate date from the publish date.

mogmarsh:
### Summary

Fixes #158
Fixes https://github.com/alleyinteractive/wp-curate/issues/169

This pull request introduces enhanced "Order By" options to the query block, providing editors with granular control over how posts are ordered, including the ability to sort by custom meta keys. The update mirrors the ordering capabilities of the core query loop block and ensures the experience is consistent for users.

#### Key Enhancements

- Adds new `order` and `orderby` options, allowing posts to be sorted by a variety of fields, including custom meta keys.
- Editors can select sort direction (`asc` or `desc`) and specify a meta key when ordering by meta values.
- Updates the QueryControls UI and logic to support these new options.
- Extends the REST API endpoint to accept `orderby`, `order`, and `meta_key` parameters.
- Introduces a filter hook (`wp_curate_order_by_meta_keys`) for extensibility, allowing themes and plugins to provide custom meta keys for ordering.
- Updates TypeScript interfaces and props throughout the codebase to support the new features.
- Updates plugin version to 2.6.4 and documents changes in the changelog.

### Related Issues

- [Add "Order By" options](https://github.com/alleyinteractive/wp-curate/issues/158)
- [Support ordering by meta key](https://github.com/alleyinteractive/wp-curate/issues/169)

### Testing Instructions

1. Insert a query block in the editor.
2. Verify that the "Order By" options, including custom meta keys, are available.
3. Select different ordering options and confirm that the posts are ordered accordingly.
4. Change the sort direction and, if ordering by meta value, select or clear the meta key.
5. Ensure the experience is consistent with the core query loop block.

### Screenshots (if applicable)

_If you made visual changes, please include screenshots here._

### Checklist

- [ ] I have tested these changes locally.
- [ ] I have added necessary documentation (if applicable).
- [ ] I have checked that all acceptance criteria from the related issue are met.
- [ ] I have reviewed the code for security and performance concerns.

### Additional Notes

- These enhancements are fully backward compatible.
- Plugin version updated to 2.6.4.
- The new filter hook allows for future extensibility regarding available meta keys for ordering.
- No breaking changes expected; please report any unexpected behavior when using advanced ordering options.


blocks/query/block.json
```diff
@@ -48,14 +48,14 @@
...
-      "default": [],
+      "default": [ ],
       "items": {
         "type": "string"
       },
       "type": "array"
     },
     "posts": {
-      "default": [],
+      "default": [ ],
```
mboynes: these are odd styling changes, I'm guessing a linter did this, but by which coding standard?

blocks/query/block.json
```diff
@@ -99,21 +99,46 @@
...
+        "author",
+        "title",
+        "name",
+        "type",
         "date",
+        "modified",
+        "parent",
+        "rand",
+        "comment_count",
+        "relevance",
```
mboynes: this is only, a-hem, _relevant_ if the query is a search

blocks/query/block.json
```diff
@@ -99,21 +99,46 @@
...
       "enum": [
+        "ID",
+        "author",
+        "title",
+        "name",
+        "type",
         "date",
+        "modified",
+        "parent",
+        "rand",
```
mboynes: rand shouldn't be an option IMO. it's heavy and useless

blocks/query/block.json
```diff
@@ -99,21 +99,46 @@
...
+        "comment_count",
+        "relevance",
+        "menu_order",
+        "meta_value",
+        "meta_value_num",
         "trending"
```
mboynes: I get that we just included all options, but let's be real, no user is going to pick many of these, so we're just adding noise. IMO the list should be contained to:

- date
- title
- meta value
- trending

Then it should be filtered, and we let real world use drive adding any others.

blocks/query/block.json
```diff
@@ -99,21 +99,46 @@
...
+        "type",
         "date",
+        "modified",
+        "parent",
+        "rand",
+        "comment_count",
+        "relevance",
+        "menu_order",
+        "meta_value",
+        "meta_value_num",
```
mboynes: generally speaking, ordering by meta is going to slow the query down (potentially by A LOT on sites that don't have a meta_value index like sites on VIP do), but `meta_value_num` can be slow even on sites with an index. how slow depends on cardinality, so I can't speak in absolutes here, but given that this is a general use plugin I think including this by default is a bad idea.

components/QueryControls/index.tsx
```diff
@@ -194,7 +286,7 @@ export default function QueryControls({
           />
           {allowedTaxonomies.map((taxonomy) => (
             <Fragment key={taxonomy.slug}>
-              { /* @ts-ignore */ }
+              { /* @ts-ignore */}
```
mboynes: another strange spacing change. at least let's do this for symmetry:

```suggestion
              {/* @ts-ignore */}
```

mboynes (approved):
Made a bunch of suggestions for your consideration, none blocking :+1:

PR #383, opened 2025-09-26
Issue-275: Mismatch between numberOfPosts attribute and size of Posts attribute can cause posts to disappear in the admin

### LINKED ISSUE #275: Mismatch between numberOfPosts attribute and size of Posts attribute can cause posts to disappear in the admin
### Description of the bug

If the `numberOfPosts` attribute is set to `3` but the `posts` attribute is `[123,456,789,null]` (4 slots), a post will get backfilled into the 4th slot, but not show up in the editor. It may show up on the front end in a different spot. This bug does not seem to affect the front end.

### Steps To Reproduce

add a query block, manually change the attributes so there is a mismatch (there may be a way to do it through the GUI, but I don't know if there is). 
Add an additional query block
Watch the requests to the rest API, you will see 4 posts are requested, but the 4th one will never appear in the editor.
Fix the mismatch, you should see the missing post.

### Additional Information

_No response_

nikkifurls:
## Summary

Fixes #275

This pull request addresses the issue where a mismatch between the `numberOfPosts` attribute and the size of the `posts` attribute can cause posts to disappear in the admin editor. Previously, if `numberOfPosts` was set to 3 but `posts` contained 4 items (e.g., `[123,456,789,null]`), a post would be backfilled into the 4th slot and not appear in the editor, although it may show up on the front end. This PR ensures consistency between these attributes to prevent such discrepancies.

## Related Issue

- [Mismatch between numberOfPosts attribute and size of Posts attribute can cause posts to disappear in the admin (#275)](https://github.com/alleyinteractive/wp-curate/issues/275)

## Steps to Reproduce

1. Add a query block and manually change the attributes to create a mismatch between `numberOfPosts` and the length of `posts`.
2. Add an additional query block.
3. Observe the REST API requests; four posts are requested, but the fourth does not appear in the editor.
4. Fix the mismatch and observe that the missing post reappears.

## Testing Instructions

- Verify that when the `numberOfPosts` and `posts` attributes are mismatched, the editor now accurately reflects all posts.
- Confirm that no posts disappear in the admin editor due to this mismatch.
- Ensure that the fix does not affect front-end rendering.

## Additional Information

_No response_


mboynes (approved):
💯

PR #387, opened 2025-09-29
Issue-378: Add option to \"Pin this Post\" in addition to \"Pin a Post\"

### LINKED ISSUE #378: Add option to "Pin this Post" in addition to "Pin a Post"
### Description

I would like to have the option to "Pin this Post" as a shortcut to pinning a backfilled post so that I don't have to search for a post in the PostPicker when it's already displayed in the editor.

### Use Case

When a user sees the post they want to pin in the backfill, it's easier to "Pin this Post" instead of searching for it in the PostPicker.

nikkifurls:
### Summary

Fixes #378

### Description

Adds the option to "Pin this Post" as a shortcut to pinning a backfilled post directly from the editor, eliminating the need to search for the post in the PostPicker when it is already displayed.

### Use Case

When a user sees the post they want to pin in the backfill, it's easier to "Pin this Post" instead of searching for it in the PostPicker.

### Testing Instructions

- Ensure that the "Pin this Post" option appears for backfilled posts displayed in the editor.
- Verify that selecting "Pin this Post" successfully pins the post without needing to use the PostPicker.
- Check that existing "Pin a Post" functionality is unaffected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to `WP Curate` will be documented in this file.
 
+## 2.7.1 - 2025-10-01
+
+- Enhancement: Add option to "Pin This Post" so editors can pin the selected post.
+
 ## 2.7.0 - 2025-09-05
 
 - Enhancement: Update wp-type-extensions to 4.0.0.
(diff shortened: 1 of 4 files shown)


samhermes (approved):
🌠

PR #396, opened 2025-10-08
TECH-208: wp-curate issue #51: Add support for selecting a layout

mogmarsh:
## Summary

wp-curate issue #51: Add support for selecting a layout

This PR introduces comprehensive support for selecting layouts and block patterns in the `wp-curate/query` block, aligning it with the functionality found in the Columns and Query Loop blocks. The implementation includes both JavaScript/React and PHP enhancements to enable variation and pattern selection, as well as improved extensibility for future layouts.

## Notes for reviewers

- The block now supports a `layout` property, enabling users to choose among different layout options (with "list" as the default).
- Pattern and variation selection modals have been integrated, allowing users to initialize blocks with custom content or layouts.
- Block variations and block patterns are now registered via both JavaScript (block.json, `registerBlockVariation()`) and new PHP classes.
- Utility functions and TypeScript types have been introduced to support pattern searching, selection, and transformation.
- The main plugin initialization now includes new features for block variations and block patterns.
- The PR includes dependency upgrades and some minor code cleanup and refactoring.

### Added

- Support for selecting a layout in wp-curate, following the pattern used in the Columns and Query Loop blocks.
- A `layout` property to the query block schema, allowing selection of different layouts (defaulting to "list").
- Ability to register block variations via block.json, `registerBlockVariation()`, or PHP filter `get_block_type_variations`.
- Integration of a variation picker modal (using `__experimentalBlockVariationPicker` or a custom component) for layout selection.
- Pattern selection modal and placeholder React components for enhanced block initialization.
- Utility functions and hooks for pattern searching and block transformation.
- New TypeScript types for block and block patterns.
- New PHP feature classes for registering block patterns and block variations.
- Initialization of block variation and pattern features in the main plugin entrypoint.

### Changed

- Updated import source for `PluginDocumentSettingPanel` reflecting changes in WordPress package structure.
- Upgraded various `@wordpress/*` dependencies to the latest versions.

### Deprecated

_None._

### Removed

- Hardcoded InnerBlocks template for the Query block, replaced by a more flexible pattern-based approach.

### Fixed

- Improved error handling and preview display during pattern and layout selection.
- Linting and code cleanup.

### Security

_None._

## Ticket(s)

- [TECH-208](https://alleyinteractive.atlassian.net/browse/TECH-208)

[...]

PR #401, opened 2025-10-14
Issue-400: Allow Query Block Template to be Filterable

### LINKED ISSUE #400: Allow Query Block Template to be Filterable
### Description

I would like to see the query block template be made filterable so it can be customized.

### Use Case

As a developer I would like to have the ability to filter the InnerBlocks template that is used by a query block.

efuller:
### Summary

Fixes #400

This pull request addresses the issue described in [#400](https://github.com/alleyinteractive/wp-curate/issues/400): "Allow Query Block Template to be Filterable". It makes the query block template filterable so it can be customized by developers.

### Proposed Changes

- Makes the InnerBlocks template used by the query block filterable.
- Enables customization of the query block template via filters.

### Testing Instructions

1. Apply the changes.
2. Use a filter to modify the InnerBlocks template of the query block.
3. Confirm that the template is updated as expected.

### Additional Notes

- No breaking changes expected.
- No new dependencies introduced.

### Checklist

- [ ] Tests added or updated
- [ ] Documentation updated (if necessary)
- [ ] Code follows project standards


mogmarsh:
@efuller heads up that https://github.com/alleyinteractive/wp-curate/pull/396 gets rid of the template, and replaces it with Block Variations and Block Templates. We should talk through making sure this need is still met.

efuller:
@mogmarsh Cool. I'll pause my work here then. Let's connect to talk it through. My initial thought is that it shouldn't be an issue and will actually work to our benefit.

efuller:
I'm going to close this out due to the template being removed with the latest or upcoming changes. We can revisit if needed.

PR #402, opened 2025-10-15
Add text align support to post title block

### LINKED ISSUE #344: Add support for `textAlign` in the WP Curate Post Title block
### Description

We have a hard requirement to be able to align post titles via the `textAlign` supports option.

**Note for clarity**: The `core/post-title` block includes support for both `align` and `textAlign`, along with other formatting options. This request is for the ability to align the **text**, though we may want to consider support for all options to maintain parity with `core/post-title`. We will need to implement these changes in the `wp-curate/post-title` block to mirror what is supported in the `core/post-title` block.

The following [block.json property](https://developer.wordpress.org/block-editor/reference-guides/block-api/block-supports/#typography-textalign) is available as of WP 6.6:

```json
"supports": {
  "typography": {
    "textAlign": true
  }
}
```

### Use Case

For certain layouts, such as a full-width image followed by a post title, we need to allow editors the option to center-align the text.

kaitlinbolling:
Adds text alignment support to the custom post title block.
Fixes #344

### DIFF
--- a/blocks/post-title/block.json
+++ b/blocks/post-title/block.json
@@ -23,6 +23,9 @@
   },
   "render": "file:render.php",
   "supports": {
-    "inserter": false
+    "inserter": false,
+    "typography": {
+      "textAlign": true
+    }
   }
 }


cahdeemer (approved):
🍁

PR #405, opened 2025-10-20
Adds notice compoment and updates language for error messsage

willowCeleste:
This PR cleans up the error/warning triggered when the Query block returns no posts:

- The message was displayed in an unstyled paragraph tag so I replaced that with the core `Notice` component which looks a lot cleaner.
- Updated the language of the error message to be a bit clearer and more specific.


mogmarsh (approved):
👍

PR #407, opened 2025-10-23
Upgrade to PHPStan 2.0, upgrade tests

srtfisher:
- Upgrade PHPStan to 2.0.
- Add `wp-graphql/wp-graphql` as a dev dependency for PHPStan.
- Drop `simplehtmldom/simplehtmldom` and replace it with Mantle's HTML class.


src/meta.php
mogmarsh, line 1: can probably remove `config/post-meta.json` as well.

mogmarsh (approved):
👍

PR #410, opened 2025-10-24
Adding tests for existing functionality

### LINKED ISSUE #168: Add tests for front end query modifications
### Description

I would like to see tests added so we can make changes to the php/front end logic with more confidence.

### Use Case

Tests will give us confidence that code changes are not causing any unexpected changes in output.

srtfisher:
Resolves #168


anubisthejackle (approved):
🎸

PR #446, opened 2026-03-14
Issue 445: Fix postType propagation

### LINKED ISSUE #445: Fix entity postType propagation
### Description of the bug

Make the post type used by `wp-curate/post-title`'s `useEntityProp` match the actual entity type, so posts render correctly without `core/post-template`.

The `wp-curate/post` block already fetches the post object via `usePostById`, which includes the resolved `type` field. It just doesn't pass it to children.

---

When attempting to use the `wp-curate/post` block within a query with no inner `core/post-template` block, I ran into an issue with posts not being returned. The project I'm working on uses an `article` post type, and `wp-curate/post-title` blocks were never being rendered.

Technical explanation from Claude:
> `wp-curate/post-title` calls `useEntityProp('postType', postType, ...)` where `postType` always resolves to `'post'` (hardcoded in the query context). Without `core/post-template`, entities are stored in the data store under their actual type (e.g., `article`), so the lookup under `post` finds nothing.

The fix is to add `postType` to the `BlockContextProvider`, add `"postType"` to the `wp-curate/post-title` `usesContext` array, and update `wp-curate/post-title` to read `postType` from direct context first, falling back to `query.postType`.

### Steps To Reproduce

On a site that does not use the 'post' `postType` (e.g., 'article'), add the following:

```html
<!-- wp:wp-curate/query {"numberOfPosts":1,"postTypes":["article"],"posts":[null]} -->
<div class="wp-block-wp-curate-query"><!-- wp:wp-curate/post -->
<div class="wp-block-wp-curate-post"><!-- wp:wp-curate/post-title /--></div>
<!-- /wp:wp-curate/post --></div>
<!-- /wp:wp-curate/query -->
```
Notice the Post Title is never rendered

### Additional Information

I have a fix for which I'll open a pull request. The commits are based on another branch and so not suitable for a straight merge.

This will change the behavior for all consumers, but should be safe since it only adds data that was previously missing.

goodguyry:
Fixes #445 

- Adds `postType` to the `BlockContextProvider`
- Adds `"postType"` to the wp-curate/post-title `usesContext` array and updates the edit compnent to read `postType` from context first, falling back to `query.postType`.
- Replaces hard-coded post type in Query an deduplication

### DIFF
--- a/blocks/post-title/block.json
+++ b/blocks/post-title/block.json
@@ -10,7 +10,7 @@
   "textdomain": "wp-curate",
   "editorScript": "file:index.ts",
   "style": "wp-block-post-title",
-  "usesContext": ["postId", "query", "pinnedPosts", "customPostTitles"],
+  "usesContext": ["postId", "postType", "query", "pinnedPosts", "customPostTitles"],
   "attributes": {
     "level": {
       "type": "number",
--- a/blocks/post-title/edit.tsx
+++ b/blocks/post-title/edit.tsx
@@ -13,6 +13,7 @@ interface PostTitleEditProps {
   };
   context: {
     postId: number;
+    postType?: string;
     query: {
       perPage?: number;
       postType?: string;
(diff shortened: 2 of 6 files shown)


mogmarsh (approved):
👍

goodguyry:
CI failures to be addressed in #447

goodguyry:
I may have additional changes to patch in. I'm going to let this sit until I know for sure.

PR #453, opened 2026-03-25
Issue 408: Change the title of the query block to avoid confusion with WP core query block

### LINKED ISSUE #408: Relabel the Query block to Curatable Query for clarity
### Description

When core's Query Loop block is able to be inserted into a block editor context, it's confusing to see "Query Loop" alongside our "Query" block, and it makes it look like our Query block is part of core. To better differentiate the two, we should relabel ours "Curatable Query." To preserve backwards compatibility, we should only change the visible label, not the actual name of the block (stored in the DB).

Image

### Use Case

When inserting a new curatable query block, it should be obvious which one editors should use.

jessicamgoddard:
Fixes #408

### DIFF
--- a/blocks/query/block.json
+++ b/blocks/query/block.json
@@ -3,7 +3,7 @@
   "apiVersion": 2,
   "name": "wp-curate/query",
   "version": "0.1.0",
-  "title": "Query",
+  "title": "Curatable Query",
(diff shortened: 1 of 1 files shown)


jakewrfoster (approved):
LGTM 👍

PR #456, opened 2026-04-10
Adds "edit" and "view" post buttons to the post block

rpasillas:
This is a helpful addition that we on Camp FAM deemed as missing.
- Adds two new toolbar buttons to the post block's toolbar
  - Edit post, opens in new tab
  - View post, opens in new tab

### DIFF
--- a/blocks/post/edit.tsx
+++ b/blocks/post/edit.tsx
@@ -312,6 +312,7 @@ export default function Edit({
       <BlockControls>
         <ToolbarGroup>
           <ToolbarButton
+            disabled={!postId}
             icon={pencil}
             label={__('Edit post in a new tab', 'wp-curate')}
             onClick={() => {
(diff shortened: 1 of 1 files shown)


blocks/post/edit.tsx
```diff
@@ -308,6 +319,28 @@ export default function Edit({
...
       )}
     >
+      <BlockControls>
+        <ToolbarGroup>
+          <ToolbarButton
+            disabled={!postId}
+            icon={pencil}
+            label={__('Edit post in a new tab', 'wp-curate')}
+            onClick={() => {
+              window.open(`/wp-admin/post.php?post=${postId}&action=edit`, '_blank');
```
mogmarsh: we can't assume WordPress will be installed at the root of the site.
Claude recommends this:
```
import { addQueryArgs } from '@wordpress/url';

const editUrl = addQueryArgs( 'post.php', {
    post: postId,
    action: 'edit',
} );
```
I think it should work, because we're already on `post.php` just editing a different ID.

mogmarsh (approved):
one recommended tweak, but otherwise 👍

blocks/post/edit.tsx
rpasillas: Nice! I made this change.

PR #458, opened 2026-04-29
Add backfill date limit default and block setting

### LINKED ISSUE #440: Default to date-limiting backfill queries
### Description

Backfill queries, especially taxonomy-term-filtered ones, can be slow on large sites. The simplest solution is to limit such queries by date, e.g. limit results to the past 30 days. Consider defaulting wp-curate to limit backfill posts to some recent time period. Options (not mutually exclusive) include...
1. Filter the default value
2. Add a per-block setting, defaulted to some recent period but giving the option to expand it
3. Add a sitewide setting (I don't think we have any of these in wp-curate, though?)

### Use Case

We want the largest publishing sites to be able to install this and use it OOTB without performance concerns

jacobschweitzer:
Adds a default backfill date limit of 30 days with a filter. Also adds a block setting with a select box dropdown for setting the limit to 30 days, 60 days, 90 days, 6 months, 1 year, and no limit. 

Resolves #440


src/class-plugin-curated-posts.php
```diff
@@ -17,6 +17,40 @@
...
+	 */
+	public static function date_query_for_backfill( int $backfill_days ): array {
+		/**
+		 * Filters the number of days to limit backfill posts to.
+		 * Return 0 to disable the date limit. The filter receives the per-block value
+		 * so site-wide overrides can ignore it.
+		 *
+		 * @param int $backfill_days Days to limit backfill posts. 0 means no limit.
+		 */
+		$backfill_days = (int) apply_filters( 'wp_curate_backfill_days', $backfill_days );
```
manewc, line 42: If we had logic leveraging this wp_curate_backfill_days, then this will effect ALL queries? For instance, this would then ignore the `Backfill Date Limit` in the `<SelectControl>` in `components/QueryControls/index.tsx`?

If this is the case, I feel like this may bring up create some UX confusion.

manewc (approved):
The code is solid, I just had a non-blocking clarification question for consideration.

🚀

PR #459, opened 2026-04-29
FAM-3249: Allows for the WP Query block to work when inside of a resusable/pattern block.

rpasillas:
Allows for rendering of WP Curate block when its inside of a pattern.


goodguyry (approved):
🎸 lgtm!

PR #462, opened 2026-05-12
FAM-3249: Allows for the WP Query block to work when inside of a resusable/pattern block

rpasillas:
Fixes deduplication when a WP Curate block is on the same page as one that is saved as a pattern/reusable block. 

This only works if both settings, Deduplication and Unique Pinned Posts, post settings are set to true, see line: https://github.com/alleyinteractive/wp-curate/blob/develop/src/post-ids/class-pinned-in-post-content.php#L47


manewc (approved):
🌶️

rpasillas:
Now that I can run PHPStan locally, once all the failing tests pass, I will close this PR without merging to open a new one, where I separate the `npm audit fix` issues into their own PR.

rpasillas:
Closing now.

PR #463, opened 2026-05-13
FAM-3249: NPM Vulnerabilities

rpasillas:
Fixes NPM vulnerabilities that were blocking #462.

`dompurify`
`fast-uri`
`postcss`

Addresses linting and type errors as a result.


anubisthejackle (approved):
🎸

PR #464, opened 2026-05-13
FAM-3249: Fixes the path for installing wp-graphql

rpasillas:
Fixes the path for installing WP Graph QL. This came up because running PHPStan from the composer file yielded this error: 

`Path /Users/{user-dir}/broadway/www/wp-curate/wp-content/plugins/wp-curate/wp-content/plugins/wp-graphql does not exist `

And running `composer install` resulted in a new directory, `wp-content/plugins/wp-graphql` being created inside of the wp-curate plugin.


anubisthejackle (approved):
🎸

PR #465, opened 2026-05-13
FAM-3249 Reusable Block Support

rpasillas:
Fixes deduplication when a WP Curate block is on the same page as one that is saved as a pattern/reusable block. 

This only works if both settings, Deduplication and Unique Pinned Posts, post settings are set to true, see line: https://github.com/alleyinteractive/wp-curate/blob/develop/src/post-ids/class-pinned-in-post-content.php#L47

Note: This is a re-do of this PR: #462 (which was approved) that I re-did so that I could separate out the NPM Audit items (#463) and the Composer items (#464) into their own PRs.


jessicamgoddard (approved):
✨

PR #468, opened 2026-06-08
Remove filter callback parameter typehints causing PHPCS failures

### LINKED ISSUE #467: Fix issue resulting in no pagination when core/query block is used
### Description of the bug

When the core/query block is used, pagination displays in the editor but not on the frontend.

### Steps To Reproduce

1. With `wp-curate` plugin activated, add a `core/query` block to a page or post
2. Set the limit to something less than the total number of queried posts to activate pagination
3. View the frontend and see that pagination does not display

### Additional Information

_No response_

jessicamgoddard:
Fixes #467

Also, CI was failing on PHPCS `Alley.PHP.FilterCallbackTypehint.ParameterTypehint` violations in WordPress filter callbacks. The failures came from newly typehinted callback parameters in two feature classes.

- **Filter callback signatures**
  - Updated `Subquery_Block_Display::filter_pre_render_block()` to remove parameter typehints from hook callback inputs.
  - Updated `Parsely_Support::add_parsely_trending_posts_query()` to remove parameter typehints from hook callback inputs.
- **Behavioral scope**
  - Return types and callback logic are unchanged; this is a compatibility/signature correction for WordPress hook execution and coding standards.

```php
// Before
public function add_parsely_trending_posts_query( array $posts, array $args ): array

// After
public function add_parsely_trending_posts_query( $posts, $args ): array
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,11 @@ All notable changes to `WP Curate` will be documented in this file.
 
 Nothing yet.
 
-## 3.2.0 -- 2026-06-09
+## 3.2.1 - 2026-06-24
+
+- Bug fix: Fix issue preventing pagination from working in core query blocks.
+
+## 3.2.0 - 2026-06-09
 
 - Enhancement: Add "edit" and "view" buttons to the Post block toolbar.
 - Enhancement: Support wp-type-extensions 5.
--- a/src/features/class-block-index.php
+++ b/src/features/class-block-index.php
@@ -30,7 +30,7 @@ final class Block_Index implements Feature {
 	 * @param array $block The block.
 	 * @return array The modified block data.
 	 */
-	public function assign_post_index( array $block_data, array $block ): array {
+	public function assign_post_index( $block_data, $block ): array {
 		if ( 'wp-curate/query' === $block_data['blockName'] ) {
 			// Find the wp-curate/post blocks inside the query block.
 			$inner_blocks = $block_data['innerBlocks'];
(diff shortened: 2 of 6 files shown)


dlh01:
What's the scenario where `"foundRows": true` wouldn't be set on the parent query block and so passed down as context to the pagination blocks? I would have thought that the paginated query results would also not work correctly without that, but I might be misremembering how those blocks work.

@mogmarsh mentioned to me that when the team was discussing this issue, one suggestion was to get rid of the `no_found_rows` logic entirely, which might also make sense. IIRC, it's there only as a performance boost, so it doesn't need to make more trouble than it's worth.

mogmarsh:
@dlh01 The scenario is a WP Curate Query block and a core Query Block and core pagination on the same page. Imagine a landing page with curated posts at the top (WP Curate Query), and then the river of everything else below (Core Query) and then pagination (core pagination).
@samhermes discovered this bug. I'm not sure if the pagination block was inside the core query block or adjacent to it, but the pagination did not work.

jessicamgoddard:
@dlh01 and @mogmarsh  FWIW, I tested on a page without any `wp-curate` blocks, only a title and a `core/query` block and had this issue. It doesn't appear `"foundRows": true` is set by default.

dlh01:
Right, it's not true by default. The thinking at the time was to make it opt-in by having the implementer add that attribute when needed, since most query blocks don't use pagination.

jessicamgoddard:
Oh, I see what your saying. 🤔 My brain is a little foggy today due to a cold. I'll poke at this a bit more and see if I can figure out the "why."

jessicamgoddard:
`$found_rows = null` by default.

mogmarsh (approved):
👍 
v3.2.0 just got released last night, so we'll probably want to bump to 3.2.1 so we can use this fix right away. You can add that to this PR. Hit me up if you need assistance.

mogmarsh (approved):
👍

PR #471, opened 2026-07-01
Don't attempt to read post_id from Parse.ly metadata

### LINKED ISSUE #470: Parse.ly metadata lookup throws a TypeError if metadata isn't a string
### Description of the bug

When using the Parse.ly integration, as a shortcut to avoid looking up a post ID by URL, we attempt to read it from the custom metadata for sites that add it. The `metadata` property in the API response for a post is supposed to be stringified JSON, but we aren't checking that before running `json_decode`, which will throw a TypeError if anything other than a string is passed (like an array).

### Steps To Reproduce

This is difficult to reproduce, because it relies on the API returning different data than we expect (which has happened at least once in the wild).

### Additional Information

I don't think we should bother checking the `post_id` field in the custom metadata at all, because it relies on users setting that as a custom field in what is being sent to Parse.ly, which most users won't do.

kevinfodness:
## Summary

Our attempts to read `post_id` from custom metadata rely on users setting that value in the custom metadata, which most users aren't going to do, so it shouldn't be our default posture. Instead, we reverse-engineer the post ID from the URL using built-in functions, which were the fallbacks anyway, and retain the filter on the derived post ID.

This also fixes a bug where Parse.ly's API returns a value other than stringified JSON for the `metadata` key, which would throw a TypeError in the previous implementation.

Fixes #470

### DIFF
--- a/src/features/class-parsely-support.php
+++ b/src/features/class-parsely-support.php
@@ -118,15 +118,9 @@ final class Parsely_Support implements Feature {
 			}
 			$ids = array_map(
 				function ( $post ) {
-					// Check if the metadata contains post_id, if not, use the URL to get the post ID.
-					$metadata = json_decode( $post['metadata'] ?? '', true );
-					if ( is_array( $metadata ) && isset( $metadata['post_id'] ) ) {
-						$post_id = (int) $metadata['post_id'];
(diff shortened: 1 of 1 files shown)


jessicamgoddard (approved):
✨

PR #473, opened 2026-07-10
Update blocks to api v3

NicoleA:
Two blocks were updated from apiVersion: 2 to apiVersion: 3:

**query block**
  - Bumped apiVersion to 3 in block.json
  - Added a blocks.registerBlockType filter in index.tsx to replace the PHP-registered variation icons (dashicon strings) with the equivalent SVG from @wordpress/icons, which is required for icons to render correctly in the apiVersion 3 iframed  editor                                                                                                                                                                                                                                              

**post block**
  - Bumped apiVersion to 3 in block.json
  - Updated the move/drop feature to resolve the editor canvas window and document from a ref on the block element, rather than using the global window/document, so that event listeners attach to the correct frame in the iframed editor
  - The ref is passed through useBlockProps() so it is properly merged with the block's internal ref


blocks/post/edit.tsx
```diff
@@ -142,6 +142,8 @@ export default function Edit({
     }
   }
 
+  const blockRef = useRef<HTMLDivElement>(null);
```
kevinfodness: Is the block guaranteed to be a div? Would it make sense to use the more generic `HTMLElement` here?

kevinfodness (approved):
One question, otherwise 🍣

PR #476, opened 2026-09-02
defaults to date-limiting backfill queries

willowCeleste:
- defaults backfill date to 30 days in the past
- allows filtering of the default backfill date
- adds overrides on a per-block basis for the default backfill date


gratrockstar (approved):
🐸

